PR #1, opened 2018-02-05
travis integration

Mizzick:
Do not build for Travis commits

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,8 @@ deploy:
       keep-history: true
       verbose: true
       on:
-        branch: release
+        branch: master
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -17,12 +20,4 @@ deploy:
       keep-history: true
```
ameshkov: I don't think we need it for `gh-pages`

ameshkov: Other than that all good

push.sh
```diff
@@ -16,7 +16,9 @@ upload_files() {
   git push github HEAD:master
 }
 
-setup_git
-commit_files
-upload_files
-
+if [ ${TRAVIS_BRANCH} == "master" ] && [ ${TRAVIS_EVENT_TYPE} != "pull_request" ]
```
ameshkov: Фу
https://docs.travis-ci.com/user/deployment/script/

Issue #18, opened 2018-02-16
AlleBlock list - wrong URL


maciejtarmas:
Hi, thanks for adding my list with [this commit](https://github.com/AdguardTeam/FiltersRegistry/commit/7a3c01059bf83f801eb3a4d4792123e3e8224ccd), though the URL should point to https://alleblock.pl/alleblock/alleblock.txt

Could you please replace the GitHub link with the official one?

Alex-302:
Hi. Done.

PR #19, opened 2018-02-16
Replaced AlleBlock list URL

Alex-302:
Replaced AlleBlock list URL https://github.com/AdguardTeam/FiltersRegistry/issues/18


ameshkov:
@Alex-302 shouldn't you also change it in the template?

Also, every rule is preceded by a number of spaces, does it work okay?

Alex-302:
Extra spaces are removed on our side
image

PR #44, opened 2018-04-19
Update metadata.json -- EasyList China

vbagirov:
+ "recommended" tag


ameshkov:
Make sure that there're no more recommended Chinese filters.

vbagirov:
@ameshkov I did, there's none. That's the point of this change.

PR #64, opened 2018-06-09
Polskie Filtry Ciasteczkowe - added AdGuard spesific source

Alex-302:


ameshkov:
What about the main list? Wouldn't it be better to include the supplement there right away?

The old supplement list can be marked as "obsolete" after that.

ameshkov:
I mean the `! Title: Official Polish filters for AdBlock, uBlock Origin & AdGuard`.

Alex-302:
This filter is not for ads. For annoying elements only.

PR #65, opened 2018-06-09
Replaced Prebake filter source

Alex-302:
https://forum.adguard.com/index.php?threads/prebake-filter-obsolete-update-link.28879/


ameshkov:
It seems that it's not obsolete anymore:
! Last modified: 28 May 2018 14:00 UTC

Alex-302:
done. `subscriptionUrl` is also removed.

PR #66, opened 2018-06-09
Official Polish filters for AdBlock, uBlock Origin & AdGuard - supplement for AdGuard is added to the main filter

Alex-302:


filters/ThirdParty/filter_238_PolskiFiltrAdg/metadata.json
```diff
@@ -9,8 +9,8 @@
   "groupId": 7,
```
ameshkov: Fix the filter description as well, mention that it's now included into the main subscription.

@AdamWr could you please assist?

filters/ThirdParty/filter_238_PolskiFiltrAdg/template.txt
```diff
@@ -1,2 +1 @@
 !
-@include "https://raw.githubusercontent.com/MajkiIT/polish-ads-filter/master/polish-adblock-filters/adblock_adguard.txt" /exclude="../../exclusions.txt"
```
ameshkov: We need to change the file contents here (mention that it is now included into the main subscription)

Alex-302: but `filter.txt` contains `! included to the main Official Polish filters for AdBlock, uBlock Origin & AdGuard`

ameshkov: It might be replaced when the filter is being rebuilt. Not that it will happen, but anyway just in case we'd better have it

Alex-302: Got it. Moved that comment to `template.txt`

ameshkov:
@Alex-302 isn't it time?

Alex-302:
I think we can merge. @AdamWr please take a look.

PR #93, opened 2018-08-20
Update metadata.json

BlazDT:
https://github.com/liamja/Prebake

This list is no longer maintained. Also domain is not resolved when visiting http://liamja.co.uk/Prebake/


ameshkov:
@Alex-302 @BlazDT guys, you need also to:

1. Add the `obsolete` tag
2. Add "disabled" attribute (so that we stopped trying to update this list from an unsafe source)

PR #111, opened 2018-10-24
Juvander's Adblock List for Finland is obsolete

Alex-302:


ameshkov:
I doubt it's obsolete, maybe a temporary issue. Making it `disabled` would be enough.

PR #113, opened 2018-10-31
Re-sorted exclusions; replace RU AdList source URL; enabled `#?#` rules in third-party filters

Alex-302:


Alex-302:
@BlazDT please take a look.

PR #120, opened 2018-11-07
Updated ABPVN List source URL

Alex-302:


AdamWr:
Wouldn't it be better to add link from GitHub?
https://raw.githubusercontent.com/abpvn/abpvn/master/filter/abpvn.txt

https://github.com/abpvn/abpvn/

ameshkov:
@AdamWr makes sense

PR #150, opened 2019-01-03
Updated filters-compiler

Mizzick:
Some fixes in platforms configuration


ameshkov:
@Alex-302 plz merge it when @Mizzick is also online so that you could check the result and quickly roll it back if something goes wrong.

PR #158, opened 2019-01-24
list kr official filter

APG2k18:
originally list kr was an recommended kor ad filter but it was replaced with a yous list after list kr filter is removed for maintainers afk. but now that its back and maybe we can set list kr as an recommended kor ad filter.

### DIFF
--- a/filters/ThirdParty/filter_227_List-KR/metadata.json
+++ b/filters/ThirdParty/filter_227_List-KR/metadata.json
@@ -12,6 +12,7 @@
     "purpose:ads",
     "purpose:privacy",
     "reference:2",
-    "lang:ko"
+    "lang:ko",
+    "recommended"
   ]
 }
--- a/filters/ThirdParty/filter_244_YousList/metadata.json
+++ b/filters/ThirdParty/filter_244_YousList/metadata.json
@@ -10,7 +10,6 @@
   "subscriptionUrl": "https://raw.githubusercontent.com/yous/YousList/master/youslist.txt",
   "tags": [
     "purpose:ads",
-    "lang:ko",
-    "recommended"
+    "lang:ko"
   ]
 }


APG2k18:
hai@alex-302

APG2k18:
hai
@alex-302

Alex-302:
It is not original filter and it is too early to do it. Let's wait a while.

ameshkov:
Let me please outline how we will proceed:

1. First, we will add the filter to the new versions of AdGuard browser extension, and AdGuard for Android. It is already available in AdGuard for Windows and Mac.
2. Then we will wait for some time and see if there are any complaints on it.
3. Then we might make it recommended if the users feedback is positive.

APG2k18:
hmm

ameshkov:
@APG2k18 huh, sorry for the delay, the ball is still on our side as we have not yet released the new versions of either AdGuard (they're all in the beta or nightly testing stage).

Again, sorry for this delay. We are going to update the browser extension in a week or two, just two major issues left to fix.

APG2k18:
still unapproved umm but time heals all wounds..? owo

ameshkov:
@APG2k18 hi! well, we have just published the updated extension version so people might start to use List-KR, it's no more listed as obsolete (even tho it's not marked as recommended yet).

So we are now at point 2 of my plan -- wait and see if there are complaints on it.

Btw, you can reach me directly on Telegram or in our Telegram chat (https://t.me/adguard_en) if you have any questions.

APG2k18:
thats a long period :O

ameshkov:
> thats a long period :O

Well, yeah, it is. It's just choosing the default filter list for a language is important.

How it goes btw, are there more people reporting issues since the browser extension release?

APG2k18:
well i dont know about this.. but im using this filter very usefully.

APG2k18:
@List-KR @ptrkr

ptrkr:
> Well, yeah, it is. It's just choosing the default filter list for a language is important.
> 
> How it goes btw, are there more people reporting issues since the browser extension release?

@ameshkov I didn't know there was an issue like this.. No further problems have been raised since the filter update.

ameshkov:
@ptrkr uh, for some reason I thought that @APG2k18 is the list maintainer, but now I realized this is you :)

So there are some questions then:

1. Do you want us to make `List-KR` the default list for Korea?
2. Does it make sense to use List-KR alongside YousList?

ptrkr:
> @ptrkr uh, for some reason I thought that @APG2k18 is the list maintainer, but now I realized this is you :)
> 
> So there are some questions then:
> 
> 1. Do you want us to make `List-KR` the default list for Korea?
> 2. Does it make sense to use List-KR alongside YousList?

Uh.. APG2k18 is not a maintainer to the List-KR filter. anyway,

1. Yeah, I want it when it is possible.
2. Both the filters that SlowMemory had developed in the past and the filter now are based on YousList, but I'll assume that is not good to use more than one filter with the same rules.

ameshkov:
@ptrkr 

Got it, thank you for the answers!

> Both the filters that SlowMemory had developed in the past and the filter now are based on YousList, but I'll assume that is not good to use more than one filter with the same or similar rules.

Duplicate rules will be ignored.

But anyway, I'll review the code of the applications to check what will happen if we change the recommended language-specific filter list. It might be that it will enable both of them.

ptrkr:
btw when will review have been completed :question:

ameshkov:
I have confirmed that if we change the recommended language-specific filter now, existing users will end up with TWO filter lists enabled: YousList and List-KR.

@ptrkr will it be a problem for you?

ptrkr:
> I have confirmed that if we change the recommended language-specific filter now, existing users will end up with TWO filter lists enabled: YousList and List-KR. New users will have only one of them enabled.
> 
> @ptrkr will it be a problem for you?

Nope. It doesn't make any problem i think. there were people who used to use both List-KR and YousList when SlowMemory's List-KR was the default korean adblock filter list.

ptrkr:
> I have confirmed that if we change the recommended language-specific filter now, existing users will end up with TWO filter lists enabled: YousList and List-KR. New users will have only one of them enabled.
> 
> @ptrkr will it be a problem for you?

Nope. It doesn't make any problem i think. there were people who used to use both List-KR and YousList when SlowMemory's List-KR was the default korean adblock filter list.

@ameshkov

ameshkov:
@ptrkr got it, thank you! 

And the last thing to check: 

After I merge this PR, it will be automatically applied to all instances of AdGuard for Windows, Mac, iOS, and Android. Browser extension will catch up after release v3.1. If you need a license for any of these, please contact me by email: `am at adguard.com`

ptrkr:
@ameshkov I see. thanks!

ameshkov:
> It seems that AdGuard for iOS/Android doesn't provide a way to select which filters to use

It actually does provide a way, but List-KR is the current default for the Korean locale:
telegram-cloud-photo-size-4-5834452230118422725-y

piquark6046:
https://github.com/AdguardTeam/FiltersRegistry/pull/158#issuecomment-1056653494

The changes of Youslist are applied at https://github.com/AdguardTeam/FiltersRegistry/commit/86ebc6cc6d61a3b566f3be7ce3c26b25587b14c5.
Please reduce interval of filter updates.

yous:
Ah, I found it. There is no '〉' sign on Safari protection entry, so I didn't think it's clickable. Please consider adding '〉' sign on the right of toggle button.

PR #162, opened 2019-02-11
Fixing download link and adding exclusions from our French filter

BlazDT:
https://forum.adguard.com/index.php?threads/up-4-net.30856/


AdamWr:
I can't find `1fichier.com##a[style]` in `Liste FR`, so is that exception still needed?
If not, maybe we should just remove it from our French filter?

BlazDT:
Looks like they removed the domain from their big rule with `#a[style]` meanwhile. Will remove it from our French filter.

PR #165, opened 2019-02-12
Update filter.txt

TheHasagi:
Deleted ||cifraclub.com.br^ (website is safe)


ameshkov:
@TheHasagi you should delete it from the source filter, otherwise it will be added back to the SDN filter automatically.

Check source filters here:
https://github.com/AdguardTeam/AdGuardSDNSFilter

PR #171, opened 2019-02-27
Safari filter rework

Alex-302:
Included Safari filter to Base filter.
Safari filter marked as obsolete.

https://github.com/AdguardTeam/AdguardFilters/pull/30207

### DIFF
--- a/filters/filter_12_Safari/metadata.json
+++ b/filters/filter_12_Safari/metadata.json
@@ -14,6 +14,7 @@
     "reference:2",
     "platform:ios",
   	"platform:ext_safari",
-  	"platform:ext_android_cb"
+  	"platform:ext_android_cb",
+    "obsolete"
   ]
 }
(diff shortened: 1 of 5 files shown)


AdamWr:
Should we add `(Obsolete)` to filter's `"name":` or in this case it's not necessary?

ameshkov:
> Should we add `(Obsolete)` to filter's `"name":` or in this case it's not necessary?

I think in this case it's not necessary

AdamWr:
Ok.

Mizzick:
@Alex-302 seems like something is wrong with 
`https://raw.githubusercontent.com/AdguardTeam/AdguardFilters/master/EnglishFilter/sections/content_blocker.txt`

Alex-302:
@Mizzick it is in another branch in `AdguardFilters` - `https://github.com/AdguardTeam/AdguardFilters/blob/safari/EnglishFilter/sections/content_blocker.txt`

PR #176, opened 2019-03-11
Added AG_onPropertyRead/Write

ameshkov:
1. Added `general_js_api.js` file with the non-minified functions implementations.
2. Changed `AG_abortInlineScript` implementation -- added `debug` parameter. If it is specified, the script will print a line to the browser console when the work is done.  
3. Added `AG_abortOnPropertyWrite` and `AG_abortOnPropertyRead` functions.
4. **`debug`** is not supposed to be used in the filter lists. It is useful for debugging purposes though.

### DIFF
--- a/filters/general_js_api.js
+++ b/filters/general_js_api.js
@@ -1,0 +1,229 @@
+// Aborts execution of a script when it attempts to write the specified property.
+// Based on AG_defineProperty (https://github.com/AdguardTeam/deep-override)
+//
+// Examples:
+// AG_abortOnPropertyWrite('String.fromCharCode');
+//
+// @param property property or properties chain
+// @param debug optional, if true - we will print warning when script is aborted.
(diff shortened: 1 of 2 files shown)


AdamWr:
In some cases with extension if `AG_abortInlineScript` is executed before `AG_abortOnPropertyWrite/Read` then `AG_abortOnPropertyWrite/Read` not works.
For example on this website - http://www.gazeta.pl/0,0.html
I added these rules to user filter:
`gazeta.pl#%#AG_abortOnPropertyWrite('pp_gemius_event');`
`gazeta.pl#%#AG_abortInlineScript(/adviewDFP/, 'Math.random');`
both of them works, but if I change the order of these rules to:
`gazeta.pl#%#AG_abortInlineScript(/adviewDFP/, 'Math.random');`
`gazeta.pl#%#AG_abortOnPropertyWrite('pp_gemius_event');`
second rule doesn't work.

PR #177, opened 2019-03-17
dropbase.com https://forum.adguard.com/index.php?threads/dropbbase-com-incorrect-blocking.31248/

BlazDT:
Fixing broken site

https://forum.adguard.com/index.php?threads/dropbbase-com-incorrect-blocking.31248/

### DIFF
--- a/filters/ThirdParty/filter_210_Spam404/template.txt
+++ b/filters/ThirdParty/filter_210_Spam404/template.txt
@@ -5,3 +5,5 @@
 @@||sentry.mba^$~third-party
 ! https://forum.adguard.com/index.php?threads/nulled-to-incorrect-blocking.30024
 @@||nulled.to^$domain=nulled.to
+! https://forum.adguard.com/index.php?threads/dropbbase-com-incorrect-blocking.31248/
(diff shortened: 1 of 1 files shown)


filters/ThirdParty/filter_210_Spam404/template.txt
```diff
@@ -5,3 +5,5 @@
 @@||sentry.mba^$~third-party
 ! https://forum.adguard.com/index.php?threads/nulled-to-incorrect-blocking.30024
 @@||nulled.to^$domain=nulled.to
+! https://forum.adguard.com/index.php?threads/dropbbase-com-incorrect-blocking.31248/
+@@||dropbbase.com^$domain=dropbbase.com
```
Alex-302: I am not sure, if we should to fix this issue. It is expected result for this filter.

BlazDT: Users can download items from it (and a text file is not spam all the time imho), so I do not see it as a Spam.

Alex-302: ok

PR #178, opened 2019-03-20
Fixing broken player

BlazDT:
https://github.com/AdguardTeam/AdguardFilters/issues/31239


Alex-302:
Is it required for browser extension also?

dnmTX:
As i am the only one who reported that problem,also hard to reproduced by any of you guys,also taking into account that it's a porn site and so far no one came forward to discuss it and i don't use the AG extension because i have the full version.I guess,you guys need to take some leap of faith on this one.
The rule blocks harmless HTML request which will not allow any video to load(on my end):
https://member.hdzog.com/sn4diyux.php   -go visit and see for yourself(it's safe)

ameshkov:
Guys, let's do something with this PR already:)

AdamWr:
Do you mean to merge it?

It seems that it happens only in Slimjet, in other browsers I can't reproduce it.

ameshkov:
Can we replace this rule with something that hides ads and not breaks anything?

AdamWr:
Ok, merged it.
This rule is from EasyList and to be honest I don't have idea what it should blocks.

PR #188, opened 2019-04-06
fix #32056 poczta.wp.pl

BlazDT:
https://github.com/AdguardTeam/AdguardFilters/issues/32056

Rule is applied to poczta.wp.pl as it is applied to wp.pl (~poczta.wp.pl gets ignored)


ameshkov:
Guys can we report this to filters authors?

AdamWr:
It was fixed in Polish filter.
Thank you.

PR #189, opened 2019-04-06
fix ##32001 mynet.com

BlazDT:
https://github.com/AdguardTeam/AdguardFilters/issues/32001


ameshkov:
Can we also report this to the filters authors?

AdamWr:
It was fixed in Polish filter.

PR #192, opened 2019-04-17
pcgames.de mobile - fixing click on main page

BlazDT:
pcgames.de mobile - fixing click on main page
Without rule the click on a link is not doing anything on main page.


Alex-302:
better report to the author

PR #194, opened 2019-05-01
Fix/windowscentral images

ameshkov:
Guys, this is a quick fix for no images in articles.
See the example here:
https://www.windowscentral.com/best-laptop-touchpad

Plz report this to easylist and merge afterwards.


ameshkov:
Ops, my bad, i'll recreate

PR #228, opened 2019-09-19
Fix fox-fan.tv 

Alex-302:
https://github.com/AdguardTeam/AdguardFilters/issues/40624

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -48,6 +48,8 @@ $popup,~third-party
 !***  `exclusion`
 !
 !### Easylist ###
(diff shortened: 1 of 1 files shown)


filters/exclusions.txt
```diff
@@ -48,6 +48,8 @@ $popup,~third-party
 !***  `exclusion`
 !
 !### Easylist ###
+! Excluding ##.adsbox as it is used too often by adblock detectors
```
AdamWr: I guess it should be `.myTestAd` instead of `.adsbox` :)

PR #230, opened 2019-10-04
Fix #41309 liveam.tv

zloyden:
https://github.com/AdguardTeam/AdguardFilters/issues/41309

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -133,6 +133,8 @@ $popup,domain=aflamsexnek.com|
 /click/*$subdocument
 =adblock_
 ://partner.$third-party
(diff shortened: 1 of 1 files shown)


filters/exclusions.txt
```diff
@@ -133,6 +133,8 @@ $popup,domain=aflamsexnek.com|
 /click/*$subdocument
 =adblock_
 ://partner.$third-party
+! https://github.com/AdguardTeam/AdguardFilters/issues/41309
+||out.pladform.ru^
```
Alex-302: use regex, because all rules which contains this part will be excluded

Alex-302: https://regex101.com/r/bAhFcm/1

Issue #235, opened 2019-10-18
Consider making AdBlockID recommended (Indonesian)


Alex-302:
https://github.com/realodix/AdBlockID

ameshkov:
Nice! What Indonesian filter list do we have as recommended by default at the moment?

Should we consider changing it to this?

Alex-302:
At the moment ABPindo is recommended.
@realodix says that AdBlockID is a separate filter https://github.com/realodix/AdBlockID/issues/45#issuecomment-543141189

> Should we consider changing it to this?

I think we can do it after some time.

ameshkov:
Well, let's add it, and then rename this issue in "Consider making AdBlockID recommended".

Alex-302:
ok

krystian3w:
https://github.com/AdguardTeam/AdguardFilters/issues/53093

Alex-302:
@3xploiton3 What can you say about it as Indonesian user? Should we make it recommended?

3xploiton3:
> @3xploiton3 What can you say about it as Indonesian user? Should we make it recommended?

absolutly yes, Highly recommended,

they are very active to maintain filters

PR #236, opened 2019-10-18
Fix AdBlockID #235

Alex-302:
https://github.com/AdguardTeam/FiltersRegistry/issues/235


filters/ThirdParty/filter_120_AdBlockID/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+  "filterId": 102,
```
AdamWr: Shouldn't it be `"filterId": 120` instead of `102` or not?

Alex-302: sure, 120. Fixed.

Issue #216, opened 2019-08-02
Mobile Ads filter list by YanFung


ameshkov:
@loveqqzj commented on [Thu Aug 01 2019](https://github.com/AdguardTeam/AdguardForiOS/issues/1157)

Hope to add a chinese filiter, it's very useful for many chinese webpages, his github homepage is https://github.com/YanFung/Ads.
The filiter address is https://raw.githubusercontent.com/YanFung/Ads/master/Mobile.txt
Please help to add it, thanks very much!

ameshkov:
Let's observe it for some time

ameshkov:
Seems good, let's replace one of the obsolete filter lists

Alex-302:
https://github.com/AdguardTeam/FiltersRegistry/pull/238

PR #238, opened 2019-10-18
Added YanFung mobile ads filter #216

Alex-302:
https://github.com/AdguardTeam/FiltersRegistry/issues/216
replaced ChinaList+EasyList(last update in 2014)


Alex-302:
@AdguardTeam/filters-maintainers

AdamWr:
I'm not sure if this list is obsolete, for example it contains rules with `$rewrite=abp-resource:` and if I'm not wrong, Adblock Plus supports this kind of rules from this year - https://adblockplus.org/releases/adblock-plus-35-for-chrome-firefox-and-opera-released

But as I recall, for some time this list was problematic (it was not available often), so maybe it's okay to remove it.

Alex-302:
> not sure if this list is obsolete

ok, let's compare in a few days
https://gist.github.com/Alex-302/13e4dd5505fa1c5fa25b61193f7f5a14 `[21.10.2019]`

ameshkov:
Guys, the problem seems to be that the maintainer does not change "Last modified" field.

But they do change the "Version" field (august was the last change it seems):
```
! Version: 201908202325
! Title: ChinaList+EasyList(修正)
! ChinaList and EasyList combination subscription
! Last modified: 03 Feb 2014 03:50 UTC
```

Alex-302:
Ok. What we should to do - replace another(which one? What about `Anti-Adblock Killer | Reek`?) or add new filter?

ameshkov:
Replace a different filter list, and remove the obsolete tag from the ChinaList

Alex-302:
> remove the obsolete tag from the ChinaList

https://github.com/AdguardTeam/FiltersRegistry/pull/240

PR #239, opened 2019-10-19
Fix LIste AR location

BlazDT:
Location moved from code.google.com.

### DIFF
--- a/filters/ThirdParty/filter_112_ListeAR/metadata.json
+++ b/filters/ThirdParty/filter_112_ListeAR/metadata.json
@@ -3,11 +3,11 @@
   "name": "Liste AR",
   "description": "Supplement for EasyList. Language: العربية",
   "timeAdded": 1404115015843,
-  "homepage": "https://code.google.com/p/liste-ar-adblock/",
+  "homepage": "https://hg.adblockplus.org/listear/",
   "expires": "2 days",
   "displayNumber": 2,
   "groupId": 7,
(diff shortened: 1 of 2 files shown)


Alex-302:
@AdguardTeam/filters-maintainers please don't miss notifications

AdamWr:
On these websites:
https://adblockplus.org/subscriptions 
https://easylist.to/pages/other-supplementary-filter-lists-and-easylist-variants.html
https://filterlists.com/lists/liste-ar
link to `Liste AR` looks like this:
https://easylist-downloads.adblockplus.org/Liste_AR.txt
so maybe we should use it too?
Just asking.

BlazDT:
That one seems to be newer according to date. Though previous website tells the new location is the one I changed to in list.

@smed79
Can you enlighten us please which is the real website and filter list url, thank you.

ameshkov:
Tbh, I think @AdamWr is right, and the proper URL for the filter list is the one on the ABP CDN.

smed79:
> Location moved from code.google.com

August 2015

> which is the real website and filter list url, thank you.

filter list url: https://easylist-downloads.adblockplus.org/Liste_AR.txt
website: https://forums.lanik.us/viewforum.php?f=98

---

FYI, **Liste AR** is supplement subscription for "**Liste_FR+EasyList**" https://issues.adblockplus.org/ticket/4890
https://git.io/vpQK1

BlazDT:
Thank you, modified source :)

PR #255, opened 2019-11-14
fix wcostream.com

BlazDT:
https://github.com/AdguardTeam/AdguardFilters/issues/44044

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -55,6 +55,8 @@ $popup,~third-party
 ! Excluding /google_ad_ because it gives a false positive on https://www.theregister.co.uk/2019/09/19/google_ad_blockers/
 ! and it is not really useful according to stats
 /google_ad_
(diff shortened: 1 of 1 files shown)


Alex-302:
why not just @@ exclusion?

BlazDT:
I see no harmful content on that path, so imho that rule is useless. Also if user only enables EasyList, it will still be broken unless adding exclusion there too.

Alex-302:
Please check with ABP. If reproduced - better report them.

AdamWr:
What about merging it or adding rule with `$badfilter` as a temporary solution until it will be fixed in `EasyList`?

AdamWr:
I reported it here - https://github.com/easylist/easylist/issues/4285 and added rule with `$badfilter`.

AdamWr:
It has been fixed in EasyList - https://github.com/easylist/easylist/commit/c149555cde19dea11605f35ac3cb6df65ace7126#diff-ce6ade6e52352499cfb347f46c329f9fL505

PR #256, opened 2019-11-18
Disable Web Annoyances Ultralist temporary

AdamWr:
Due to issue with `!#include` directive probably.


Alex-302:
@AdamWr please don't forget to enable it after fixing in filters compiler.

AdamWr:
Ok.

PR #270, opened 2019-12-13
'Fanboy's Cookiemonster List' renamed to 'EasyList Cookie List'

Alex-302:
https://github.com/easylist/easylist/blob/master/fanboy-cookie.template


AdamWr:
By the way, should we also change description in `Fanboy's Annoyances`?
https://github.com/AdguardTeam/FiltersRegistry/blob/master/filters/ThirdParty/filter_122_FanboysAnnoyances/metadata.json#L4

ameshkov:
Yeah

PR #283, opened 2019-12-26
added trust level option

tvinzz:


filters/trust-levels/exclusions-full.txt
```diff
@@ -0,0 +1,9 @@
+! Allowed rules by FULL trust level:
```
ameshkov: I think these lists should be set in the "FiltersCompiler", there's no need to make it configurable here

tvinzz: trust-level config lists removed from `FilterRegistry` and will be added to `FilersCompiler`

PR #290, opened 2020-01-07
Excluding unsupported options

BlazDT:
There is also the following options, but unsure where to put them:

$1p
$3p
$beacon
$popunder  (from uBlock)

$popunder could maybe rewritten to $popup (if it does the same)

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -10,13 +10,16 @@ replace=
 -abp-properties
 :xpath(
 (:scope
+$ping
 ! Exluding unsupported uBlock rules
 ##+js
+#@#+js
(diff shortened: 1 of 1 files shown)


Alex-302:
```
$1p
$3p
```
these option must be converted on our end.

BlazDT:
Thank you - opened topic for compiler: https://github.com/AdguardTeam/FiltersCompiler/issues/65

slavaleleka:
[FiltersCompiler#65](https://github.com/AdguardTeam/FiltersCompiler/issues/65) is done. is the PR still relevant?

ameshkov:
Seems that it's not

PR #294, opened 2020-01-15
fix h5.17k.com

BlazDT:
https://github.com/AdguardTeam/AdguardFilters/issues/47732

Since their filter dev cannot reproduce due to missing iOS (I cannot reproduce on my iPad either) and therefore is not removing the rule from wide block rule, whitelisting that rule.

### DIFF
--- a/filters/ThirdParty/filter_104_EasyListChina/template.txt
+++ b/filters/ThirdParty/filter_104_EasyListChina/template.txt
@@ -1,6 +1,9 @@
 !
 @include "https://easylist-downloads.adblockplus.org/easylistchina.txt" /exclude="../../exclusions.txt"
 !
+! https://github.com/AdguardTeam/AdguardFilters/issues/47732
+!+ PLATFORM(ios, ext_safari)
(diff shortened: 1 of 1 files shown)


filters/ThirdParty/filter_104_EasyListChina/template.txt
```diff
@@ -1,6 +1,9 @@
 !
 @include "https://easylist-downloads.adblockplus.org/easylistchina.txt" /exclude="../../exclusions.txt"
 !
+! https://github.com/AdguardTeam/AdguardFilters/issues/47732
+!+ PLATFORM(ios, ext_safari)
+@@$script,subdocument,third-party,websocket,xmlhttprequest,domain=h5.17k.com
```
Alex-302: does it exclude the necessary requests(which are blocked by Easylist China)?

BlazDT: It works fine on my iPad as mentioned (even without that rule while EasyList China is active).

Alex-302: can user check it? 
That rule must block something. You can check it in Safari console with remote debugging.

PR #298, opened 2020-02-03
Updated compiler version to 1.0.41

tvinzz:
Don't merge, we have to check the filters carefully first.


ameshkov:
```
Error: Cannot find module 'scriptlets/dist/cjs/scriptlets.js'
```

PR #305, opened 2020-02-17
Added high trust level for 221, 237, 247 filters

tvinzz:


krystian3w:
This should recover excluded CSS without hide uBO suplements.

PR #315, opened 2020-03-19
Disabled BarbBlock filter(last update: Aug 02, 2019)

Alex-302:
**do not merge**


ameshkov:
IT is not actually obsolete, it's just there was nothing to add to that filter list as of last time

PR #318, opened 2020-03-30
Updated exclusions for the wide IP rules

Alex-302:
https://github.com/AdguardTeam/AdGuardSDNSFilter/pull/281


AdamWr:
Do I understand correctly, that exclusion like this - `||51.77.` will remove all rules which contains it?
For example, in Base filter we have this rule:
```
@@||51.77.53.104/*.js$domain=unique-tutorials.info|getintopc.com|solvetube.site|tech-platform.info
```
Will it be excluded from list?
If so, then I think we should only exclude these rules which start with `||`.

Alex-302:
> Will it be excluded from list?

yes. I will convert them to regex syntax.

ameshkov:
@Alex-302 these IP rules are now different in Easylist

Alex-302:
I see...

PR #323, opened 2020-04-07
Fix #322 - Added uBO-specific source to AdBlockID filter

### LINKED ISSUE #322: AdBlockID+
https://github.com/AdguardTeam/AdguardFilters/issues/53093

Alex-302:
Also we can change `trustLevel` to `High` and make this filter recommended. It is maintained and does not causes issues.
https://github.com/AdguardTeam/FiltersRegistry/issues/235


slavaleleka:
then we'll have both **AdBlockID** and **ABPindo** as recomended for Indonesian. it may lead to some rules' conflicts, isn't it?

Alex-302:
@3xploiton3 uses both filters. It seems evewything is fine.
In any case we can make recommended AdBlockID only.

slavaleleka:
so let's make `high` and `recomended` and we'll see if any issues come up

Alex-302:
I will make new PR.

3xploiton3:
> @3xploiton3 uses both filters. It seems evewything is fine.
> In any case we can make recommended AdBlockID only.

i'm using 2 of them work very well, desktop and mobile, my family device too

PR #333, opened 2020-05-06
Added Chinese Filter template

Alex-302:
https://github.com/AdguardTeam/AdguardFilters/pull/54988


AdamWr:
Shouldn't we also include `EasyList China`, like in another language specific filters or is not needed in this case?

slavaleleka:
`EasyList China` might be included in FiltersRegistry

BlazDT:
I agree it should be based on EasyList China.

Alex-302:
Removed "Korean Adblock List"(224), changed filterId to 224

filters/filter_17_Chinese/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+  "filterId": 224,
+  "name": "AdGuard Chinese filter",
+  "description": "EasyList China + AdGuard Chinese filter. Filter list that specifically removes ads on websites in Chinese language.",
+  "timeAdded": 1588759046083,
+  "homepage": "https://kb.adguard.com/en/general/adguard-ad-filters#chinese-filter",
+  "expires": "4 days",
+  "displayNumber": 1,
+  "groupId": 7,
+  "subscriptionUrl": "https://filters.adtidy.org/extension/chromium/filters/17.txt",
```
slavaleleka, line 10: `../224.txt` should be here (as filterId value), isn't it?

filters/filter_17_Chinese/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+  "filterId": 224,
```
slavaleleka, line 2: is it still problematic to have `filterId=17`? even if it's obsolete filter replacement and not new filter adding

or the problems may happen exactly because of `filterId`?

ameshkov: I am not sure how old AdGuard versions will handle complete filter list removal so it's safer to replace a filter list instead.

filters/filter_17_Chinese/metadata.json
ameshkov: Yes

Alex-302: Changed https://github.com/AdguardTeam/FiltersRegistry/pull/333/commits/222b95eeecb343c6c31a645bf18bc4f94dbe2de6

PR #369, opened 2020-07-27
update few descriptions for zh locale

slavaleleka:


AdamWr:
Not a big issue, but it seems that these "pluses" looks different then other "pluses".
Most of pluses looks like this - `+`
But two of them looks like this - `＋`
Shouldn't all "pluses" look the same?

el-termikael:
@AdamWr you are right. thank you for noticing, now all pluses look this way: +.

zloyden:
Oh, these small `+` look much better)

PR #384, opened 2020-10-12
fix #65232 r20.rs6.net

Sergey-Lyapin:
https://github.com/AdguardTeam/AdguardFilters/issues/65232


Alex-302:
Not sure if this is required. This filter does not use exclusions, and broken sites - not an issue in this case.
@ameshkov should we fix it on our end?

PR #387, opened 2020-10-25
Remove AdBlockID Plus

realodix:
Following up on this issue https://github.com/realodix/AdBlockID/pull/637, the AdBlockID Plus link should be removed.


ameshkov:
@realodix thank you!

PR #388, opened 2020-10-30
fix #66624 kaspersky.com

Sergey-Lyapin:
https://github.com/AdguardTeam/AdguardFilters/issues/66624

The rule `/s_code_`  from Fanboy Enhanced Tracking list breaks kaspersky.com only in programs for mac, windows, android.


Alex-302:
wrote to our chat

Alex-302:
@Sergey-Lyapin

PR #449, opened 2021-03-25
Excluded Google Contributor anti-adblock baits

AdamWr:
Related to this - https://github.com/AdguardTeam/FiltersRegistry/pull/435
This rule `##[class^="div-gpt-ad"]` was added recently - https://github.com/easylist/easylist/commit/1320348c12e4718e486b10649fd819040bcd53c2
`div[class^="div-gpt-ad"]` is still in `Liste FR`, so shouldn't be removed.

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -57,6 +57,7 @@ $popup,~third-party
 /^##\.adsbox$/
 /adframe.
 ! Excluding Google Contributor anti-adblock baits
+/^##\[class\^="div-gpt-ad"\]$/
 /^##div\[class\^="div-gpt-ad"\]$/
 /^##\.div-gpt-ad$/
 ! https://github.com/easylist/easylist/commit/e2efe2c3686cf6192f4e2d6a944ebe092c06f27a


filters/exclusions.txt
```diff
@@ -57,6 +57,7 @@ $popup,~third-party
 /^##\.adsbox$/
 /adframe.
 ! Excluding Google Contributor anti-adblock baits
+/^##\[class\^="div-gpt-ad"\]$/
```
zloyden, line 60: I guess it might also help here https://github.com/AdguardTeam/AdguardFilters/issues/78397 without `gigazine.net##body > div[class][style*="bottom: 0px; left: 0px;"][style*="z-index:"][style*="background-color:"][style*="display: flex; justify-content: center;"]`. What do you think?

zloyden (approved):

filters/exclusions.txt
AdamWr: Yes, it should also fix these issues - https://github.com/AdguardTeam/AdguardFilters/issues/78400, https://github.com/AdguardTeam/AdguardFilters/issues/78390, but I think that leaving this rule is fine.
I'm also thinking about adding `||fundingchoicesmessages.google.com^$third-party` to AdGuard Annoyances filter, it should fixes a lot of cookie notice and anti-adblock issues.

Alex-302 (approved):

PR #458, opened 2021-04-12
Updated ABPPersian

Alex-302:
Updated ABPersian, removed dangerous rules

### DIFF
--- a/filters/ThirdParty/filter_222_ABPPersian/template.txt
+++ b/filters/ThirdParty/filter_222_ABPPersian/template.txt
@@ -419,7 +419,6 @@ _vps-banner*.gif
 ||cloob.com^$third-party
 ||chare.in^$third-party
 ||dabi.ir$~third-party
-||digikala.com^$third-party
 ||delearia.org^$third-party
 ||dude-club.com^$third-party
 ||bezoodi.com^$third-party
--- a/filters/ThirdParty/filter_222_ABPPersian/template.txt
+++ b/filters/ThirdParty/filter_222_ABPPersian/template.txt
@@ -8,7 +8,6 @@
 ! ||digikala.com^$third-party
 !
 ! ##*[class$="_ads"]:not(body):not(input)
-! ##*[class$="_ads"]:not(body):not(input)
 ! ##*[class^="ads-"]:not(body):not(input)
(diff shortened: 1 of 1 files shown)


Alex-302:
@AdguardTeam/filters-maintainers

filters/ThirdParty/filter_222_ABPPersian/template.txt
```diff
@@ -409,6 +419,7 @@ _vps-banner*.gif
 ||cloob.com^$third-party
 ||chare.in^$third-party
 ||dabi.ir$~third-party
+||digikala.com^$third-party
```
AdamWr, line 422: Do I understand correctly that this rule will be added to filter?
If so, then if I'm not wrong it will breaks a website from this issue - https://github.com/AdguardTeam/AdguardFilters/issues/53365

Alex-302: forgot to remove. Removed.

filters/ThirdParty/filter_222_ABPPersian/template.txt
```diff
@@ -6,16 +6,27 @@
 ! Attention: check if these rules are removed after update this filter
 ! https://github.com/AdguardTeam/AdguardFilters/issues/53365
 ! ||digikala.com^$third-party
-! https://github.com/AdguardTeam/AdguardFilters/issues/60229
-! ##*[class$="_ads"]:not(body):not(input)
-! ~flickr.com##*[class$="-ads"]:not(body):not(input)
 !
-! *web&utm_campaign
+! ##*[class$="_ads"]:not(body):not(input)
+! ##*[class$="_ads"]:not(body):not(input)
```
AdamWr, line 11: It seems that these rules are the same, but as it's just a comment then I guess it doesn't matter.

AdamWr (approved):

filters/ThirdParty/filter_222_ABPPersian/template.txt
Alex-302: Yes. The main change - disabled dangerous rules.

Alex-302: removed duplicate comment

PR #469, opened 2021-05-05
Removed bad rules from ABPPersian and xinggsf

Alex-302:
Removed bad rule


AdamWr (approved):

Alex-302:
also excluded bad rule from `xinggsf`

AdamWr (approved):

PR #472, opened 2021-05-11
AG-7790 allow removeparam modifier for extensions

maximtop:


filters/filter_17_TrackParam/metadata.json
```diff
@@ -13,11 +13,7 @@
   ],
   "platformsExcluded": [
     "ios",
-    "ext_safari",
-    "ext_android_cb",
```
Alex-302, line 17: `ext_android_cb` does not support `removeparam`

Alex-302 (changes requested):

ameshkov (approved):

filters/filter_17_TrackParam/metadata.json
maximtop: reverted

Alex-302 (approved):

PR #481, opened 2021-05-24
Exclude redundant regex in EasyList China

Yuki2718:
`/\.com\/[0-9a-z]{2}\/[0-9a-z]{2}\/[0-9a-z]{2}\/[0-9a-z]{32}\.js/$script,third-party` in EL China is made redundant by `/^https?:\/\/([a-z0-9]{8,10}\.)?[a-z0-9-]{5,}\.(com|bid|link|online|top|club)\/([a-z0-9]{2}\/){3}[a-f0-9]{32}\.js/$script,third-party` in Base so should be excluded just like done in French list.

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -220,7 +220,7 @@ _campaign=$popup
 !################
 !
 !### ChinaList+EasyList ###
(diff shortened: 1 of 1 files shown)


Alex-302:
Does it break something?

Yuki2718:
@Alex-302 No, it's simply redundant.

Alex-302:
I think no need to remove it.

Yuki2718:
> I think no need to remove it.

Despite a similar filter is excluded in French list? I don't get it.

Alex-302:
French is problematic.

PR #483, opened 2021-05-27
Excluded old BAB bait

Alex-302:
original rule is absent in Easylist, exists in Bulgarian only.
After that, elemhide exclusions could be removed.

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -44,6 +44,7 @@ $popup,~third-party
 !
 !### Anti-adblock baits ###
 ! all filters
+/###banner_ad$/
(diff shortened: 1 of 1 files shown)


filters/exclusions.txt
```diff
@@ -44,6 +44,7 @@ $popup,~third-party
 !
 !### Anti-adblock baits ###
 ! all filters
+/###banner_ad$/
```
Alex-302, line 47: witout `^`, because `~bg-ikonomika.com###banner_ad`

AdamWr (approved):

zloyden (approved):

PR #486, opened 2021-05-28
Fix https://github.com/AdguardTeam/AdguardFilters/issues/84173

Yuki2718:
Way too generic. BTW is exclusion for ABP Japanese Filter still needed?

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -259,7 +259,7 @@ _campaign=$popup
 !
 !### NoCoin Filter List ###
 ! https://github.com/AdguardTeam/AdguardFilters/issues/84173
-/^##\.ad$/
+/.*##\.ad$/
 !###############
 !
 !----- TEMPORARY -----
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -257,5 +257,10 @@ _campaign=$popup
 ||github.io^$third-party,xmlhttprequest,domain=~facebook.com|~gametactic.org|~github.com|~github.io|~tumblr.com|~wottactic.com
 !###########################
 !
+!### NoCoin Filter List ###
+! https://github.com/AdguardTeam/AdguardFilters/issues/84173
+##.ad
+!###############
+!
 !----- TEMPORARY -----
 !----- TEMPORARY -----
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -257,9 +257,9 @@ _campaign=$popup
 ||github.io^$third-party,xmlhttprequest,domain=~facebook.com|~gametactic.org|~github.com|~github.io|~tumblr.com|~wottactic.com
 !###########################
 !
-!### NoCoin Filter List ###
+!### Hungarian filter ###
 ! https://github.com/AdguardTeam/AdguardFilters/issues/84173
-/.*##\.ad$/
+/^##\.ad$/
 !###############
 !
 !----- TEMPORARY -----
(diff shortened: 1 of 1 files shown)


Alex-302:
> is exclusion for ABP Japanese Filter still needed?

No.

filters/exclusions.txt
```diff
@@ -257,5 +257,10 @@ _campaign=$popup
 ||github.io^$third-party,xmlhttprequest,domain=~facebook.com|~gametactic.org|~github.com|~github.io|~tumblr.com|~wottactic.com
 !###########################
 !
+!### NoCoin Filter List ###
+! https://github.com/AdguardTeam/AdguardFilters/issues/84173
+##.ad
```
Alex-302, line 262: better regexp, like `/##\.an-advert-banner$/`

Alex-302: without `^`

Yuki2718: How about `/^##\.ad$/`?

Yuki2718:
> 
> No.

Can I delete the part in this PR? Maybe it's better for you to do.

filters/exclusions.txt
Yuki2718: It seems `^` is not needed.

Alex-302:
> Can I delete the part in this PR? Maybe it's better for you to do.

Yes, you can delete it. But check if they are not exists in another filters. In general this is not a big issue.

filters/exclusions.txt
```diff
@@ -257,5 +257,10 @@ _campaign=$popup
 ||github.io^$third-party,xmlhttprequest,domain=~facebook.com|~gametactic.org|~github.com|~github.io|~tumblr.com|~wottactic.com
 !###########################
 !
+!### NoCoin Filter List ###
+! https://github.com/AdguardTeam/AdguardFilters/issues/84173
+/^##\.ad$/
```
Alex-302, line 262: With `^` the rules like  `~example.com##.ad` will not be excluded.
Maybe better `/~.*##\.ad$/`

Alex-302: I mean this rule: https://github.com/AdguardTeam/FiltersRegistry/blob/28ddd0b6c22c9d50eb7d6d1593eff80ea0461f62/filters/ThirdParty/filter_216_PolskiFiltr/filter.txt#L232

Yuki2718: Okay.

Yuki2718: I think `~` in the head should not be added.

Yuki2718:
There are special mention about `_campaign=$popup` exclusion of ABPJP in Hebrew, French, and RU AdList.

filters/exclusions.txt
```diff
@@ -257,5 +257,10 @@ _campaign=$popup
 ||github.io^$third-party,xmlhttprequest,domain=~facebook.com|~gametactic.org|~github.com|~github.io|~tumblr.com|~wottactic.com
 !###########################
 !
+!### NoCoin Filter List ###
+! https://github.com/AdguardTeam/AdguardFilters/issues/84173
+/.*##\.ad$/
```
Alex-302, line 262: Checked again. The rule in Polish filter is limited to PL only - `***,pl##.ad`
But hufilter really has wide rule `##.ad`. So better revert to `/^##\.ad$/`

filters/exclusions.txt
Alex-302: `/.*##\.ad$/` will exclude domain specific rules.

Yuki2718: > `/.*##\.ad$/` will exclude domain specific rules.

Oh, sure.

filters/exclusions.txt
Yuki2718: Done.

Alex-302 (approved):

PR #508, opened 2021-07-27
Temporary disable xinggsf

zloyden:
Our build fails because the domain `gitee.com` is not stable. Let's temporary disable it.


AdamWr (approved):

Alex-302:
do not forget to enable it later

PR #507, opened 2021-07-26
Disable Ru AdList filters

zloyden:
According to our new rules, these filters don't pass them.

PR #517, opened 2021-08-09
Removed RU Adlist and BitBlock filters

Alex-302:
RU Adlist as incompatible; BitBlock - in according to https://github.com/AdguardTeam/FiltersRegistry#what-filters-can-be-added-to-this-repository
Related to https://github.com/AdguardTeam/FiltersRegistry/pull/507

### DIFF
--- a/filters/ThirdParty/filter_115_RUAdList/metadata.json
+++ b/filters/ThirdParty/filter_115_RUAdList/metadata.json
@@ -11,7 +11,8 @@
   "tags": [
     "purpose:ads",
     "lang:ru",
-    "reference:101"
+    "reference:101",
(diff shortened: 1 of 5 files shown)


zloyden:
If I remember correctly, he has a position that if there are problems with our filters, then we need to turn them off instead of ways to solve the problem. The filter also does not pass our policy on the part of rules 1, 3.1, 5, 8. What about `Counters`? In this filter, it contains some cosmetic rules against yandex, 4pda, rambler etc,

zloyden (approved):

Alex-302:
> Counters

It may be useful.

Issue #546, opened 2021-11-29
Barblock


slavaleleka:
@alhambra11 commented on [Sat Nov 27 2021](https://github.com/AdguardTeam/AdguardBrowserExtension/issues/1910)

I'm running the beta version of adguard for chrome, and I see Barblock list is still here. Since it's not maintained anymore, wouldn't be better to get rid of it?

ameshkov:
Is it really not maintained or is there nothing to add?

zloyden:
Hm, last update was in 2019. I think we can consider it obsolete.

PR #548, opened 2021-11-29
Disable BarbBlock

zloyden:
https://github.com/AdguardTeam/FiltersRegistry/issues/546


slavaleleka (approved):

AdamWr:
Maybe we should also add `(Obsolete)` to the "name" like in other filters?
For example:
https://github.com/AdguardTeam/FiltersRegistry/blob/5eef63535576f749d525ec48151c344715f5ff22/filters/ThirdParty/filter_210_Spam404/metadata.json#L3

zloyden:
Oh, it was merged. In one hand, it's not necessary, because the filter won't be shown to the user, but it might be useful for our search for old filters.

Issue #566, opened 2022-01-29
For Icelandic, two filters are recommended


Einarsson1221:
For Icelandic users, "Dandelion Sprout's Nordic Filters" and "Icelandic ABP List" are recommended. These are good for one or the other, aren't they?

krystian3w:
@DandelionSprout

iam-py-test:
I have no idea. Let me fire up a VM and see if I can repro
@Einarsson1221 please provide AdGuard and browser versions

iam-py-test:
Still need screenshots to confirm what you are talking about

krystian3w:
I suppose he ask how many breakages generate turn on both.

Now we know AdGuard Russia and Ru AdList is incompatible vice-versa (one should be always disabled).

DandelionSprout:
Despite *Icelandic ABP List*'s "Last updated" date being 23th of December 2021, my understanding is that most of that list hasn't been changed since circa 2016.

On that ground, I'd say that my list is somewhat better for Icelandic sites, despite me too having reached a soft limit for how many Icelandic sites there are out there that entries can be added for.

ameshkov:
Here's a quick explanation on how "recommended" works:

AdGuard automatically enables recommended filter lists if it detects that you're visiting websites in that language. It will enable **all** recommended lists for that language.

How detection works:

1. Upon AG installation it checks your keyboard layouts and assumes that you're visiting websites in the given language if you have its layout.
2. While using AG it checks the language of the pages you're visiting (if this functionality is provided by the browser) OR it checks top-level domains of the visited websites (in this case you should visit them rather often, one domain is not enough to trigger auto-enabling of a filter list).

Einarsson1222:
@krystian3w @iam-py-test @DandelionSprout @ameshkov 

I forgot my account password, so I had to recreate my account.

My reasons for launching this issue are as follows. When I visited the Icelandic website, two filters were automatically enabled. However, I don't need the "Icelandic ABP List" if "Dandelion Sprout's Nordic Filters" are enabled. So I thought you might want to consider removing your recommendation of the "Icelandic ABP List" so that this filter is not automatically enabled.

Thanks for your comment.

iam-py-test:
> I forgot my account password, so I had to recreate my account.

Do you have a recovery email? You should be able to recover your account without your password

krystian3w:
[...]

PR #573, opened 2022-02-08
One recommended filter

zloyden:
According to the https://github.com/AdguardTeam/FiltersRegistry/issues/566#issuecomment-1031272201 and https://github.com/AdguardTeam/FiltersRegistry/issues/566#issuecomment-1031339492, I think we can leave only one recommended filter.


ameshkov (approved):

slavaleleka:
so which one will be recommended for `is` language?

zloyden:
> so which one will be recommended for is language?

`Dandelion Sprout's Nordic Filters`.

PR #586, opened 2022-03-08
Excluded old only invalid rules. Disabled obsolete filters

Alex-302:
Excluded old only invalid rules. Disabled ABP Japanese Filter, Schacks Adblock Plus liste.

Added to exclusions.txt, because there are many unnecessary lines about invalide rules.

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -267,7 +267,6 @@ _campaign=$popup
 ||www.dwrean.net##iframe[src="http://kratisinow.digitup.eu/widget/widget-artists"]
 !
 ! Xfiles
-||the-japan-news.com/modules/js/lib/fgp/fingerprint2.js$script,redirect=fingerprint2.js,important
 ||widgets.outbrain.com/outbrain.js$script,redirect=widgets.outbrain.com/outbrain.js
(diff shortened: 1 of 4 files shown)


filters/exclusions.txt
```diff
@@ -264,5 +258,30 @@ _campaign=$popup
...
+!
+!### Invalid rules ###
+! This section is for known old invalid rules
+!
+! void.gr
+||www.dwrean.net##a[href="http://www.kratisinow.gr"]
+||www.dwrean.net##iframe[src="http://kratisinow.digitup.eu/widget/widget-artists"]
+!
+! Xfiles
+||the-japan-news.com/modules/js/lib/fgp/fingerprint2.js$script,redirect=fingerprint2.js,important
```
AdamWr, line 270: If I'm not wrong, this rule should be converted into `redirect=fingerprintjs` - https://github.com/AdguardTeam/Scriptlets/issues/187
So, maybe it shouldn't be excluded?

Alex-302:
@slavaleleka alarm!

filters/exclusions.txt
Alex-302: ok

AdamWr (approved):

PR #597, opened 2022-03-23
general phrases

Panuka:
This is can be just dynamic pictures, not only ads.


Alex-302:
Better write about this to Liste-FR filter's author @smed79, with excplanation.

PR #613, opened 2022-05-13
Moved PeterLowesList to privacy group

jellizaveta:
Moved PeterLowesList to privacy group

### DIFF
--- a/filters/ThirdParty/filter_204_PeterLowesList/metadata.json
+++ b/filters/ThirdParty/filter_204_PeterLowesList/metadata.json
@@ -6,10 +6,10 @@
   "homepage": "https://pgl.yoyo.org/adservers/",
   "expires": "4 days",
   "displayNumber": 4,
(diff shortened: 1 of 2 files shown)


zloyden (approved):

slavaleleka:
shouldn't we change the description?

Alex-302 (approved):

Alex-302:
Exclusions/badfiltering from template.txt may be removed, I think so.

zloyden:
> Exclusions/badfiltering from template.txt may be removed(website breakages are not considered a problem by design), I think so.

Well, we can remove them from the template, but if the filter causes too many issues, we should consider disabling it in the future.

Alex-302:
> we should consider disabling it in the future.

I agree.

Alex-302 (approved):

zloyden (approved):

PR #627, opened 2022-06-17
Remove add email to account popup from Reddit

PolicyPuma4:

### DIFF
--- a/filters/filter_14_Annoyances/filter.txt
+++ b/filters/filter_14_Annoyances/filter.txt
@@ -38263,4 +38263,6 @@ hromadske.ua#$#body { overflow: visible !important; }
 !
 ! Social locker widget
 !
+! Remove add email to account popup from Reddit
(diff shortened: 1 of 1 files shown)


Alex-302:
They use randomized classes. Is Adguard Annonances enabled?

PolicyPuma4:
> They use randomized classes. Is Adguard Annonances enabled?

Yes, the AdGuard annoyance filter is enabled. While the class seems to be a randomly generated string, it is very much the same for every user, I've tested the filter on multiple browsing sessions and accounts and it works across them all.

Alex-302:
Ok, please show it in dev tools.
I don't know how to reproduce it, with or without account.

PolicyPuma4:
> Ok, please show it in dev tools. I don't know how to reproduce it, with or without account.

This was taken with a clean install of Chrome and I created an account on the spot.

Alex-302:
Oh, this is useful reminder:) I guess they delete not confirmed accounts.

Issue #636, opened 2022-07-03
"redirect:" in "exclusions.txt" is too broad


DandelionSprout:
`redirect:` in https://github.com/AdguardTeam/FiltersRegistry/blob/master/filters/exclusions.txt seems to cause the incorrect exclusion of `##.mob-app-redirect:not(html, body)` in Dandelion Sprout's Annoyances List, which I've had to change to simply `##.mob-app-redirect` in the meantime.

ameshkov:
@zloyden do we still need it?

zloyden:
I'm not sure that we need this.

zloyden:
@DandelionSprout you can revert your rule in 1-2 hours.

PR #638, opened 2022-07-05
`redirect:` in `exclusions.txt` is too broad

zloyden:
https://github.com/AdguardTeam/FiltersRegistry/issues/636
[Regex](https://regex101.com/r/5DGFet/1)

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -17,8 +17,7 @@ $inline-font
 ! Excluding redirect instructions
 ! $redirect= - processed by trust-levels
 ! ,redirect= - processed by trust-levels
-Redirect:
-redirect:
+/^\!\s*(R|r)edirect:/
 !
 ! Excluding foreign Extended CSS rules for Facebook
 facebook.com#?#


AdamWr:
To be honest, I don't know why it have to be excluded, because it's not supported (I'm guessing)?
By the way, what will happens if there will be two (or more) exclamation marks?
Like this:
```
!! Redirect:
!! ! Redirect:
!! !   ! Redirect:
!! !  ! ! Redirect:
```
Should we remove it too?
If so, then maybe something like this will be okay:
```
^!((?:\s*!?)+)(R|r)edirect:
```
Though, I'm not sure if it's needed.

zloyden:
> I don't know why it have to be excluded, because it's not supported

I think the filter build will fail like here https://github.com/AdguardTeam/FiltersCompiler/issues/150.

> By the way, what will happens if there will be two (or more) exclamation marks?

According to [this](https://help.eyeo.com/adblockplus/how-to-write-filters), it also has
```
! Redirect: http://example.com/list.txt
```

zloyden:
> but the question is, what will happens, if it will starts with more than one !?

Actually, it's wrong syntax and I can't check this due to the issue above.
@tvinzz will the build fail in `@AdamWr` case?

AdamWr:
We could also just remove every `[Rr]edirect:` which starts with `!` and I guess it should be okay too.
```
^!.*[Rr]edirect:
```

Alex-302 (approved):

AdamWr:
Currently we have `Redirect:` and `redirect:`, so as far as I understand, it removes everything which contains this phrase, so I don't think that removing comments (from third-party filters only, if I'm not wrong) is a problem, but okay :)

Alex-302:
> Currently we have `Redirect:` and `redirect:`....

It just hasn't caused any problems yet:)

Issue #656, opened 2022-08-16
ABPVN List version for content blockers (AdGuard Content Blocker)


ameshkov:
ABPVN recently made a version that's compatible with Samsung content blockers (i.e. it uses only the core ABP syntax):
https://github.com/abpvn/abpvn/blob/master/filter/abpvn_content_blocker.txt

We should update FiltersRegistry to use that version.

Here's what I propose:
1. Add a new file to FiltersRegistry that will look like this:
````
!#if content blocker
!#include the new source
!#else
!#include the old source
!#endif
```
2. Change `template.txt` for the filter to include that file instead of including the ABPVN filter directly.

PR #657, opened 2022-08-17
ABPVN List version for AdGuard Content Blocker

zloyden:
https://github.com/AdguardTeam/FiltersRegistry/issues/656

### DIFF
--- a/filters/ThirdParty/filter_214_ABPVNList/platforms.txt
+++ b/filters/ThirdParty/filter_214_ABPVNList/platforms.txt
@@ -1,0 +1,5 @@
+!#if (adguard_ext_android_cb)
+!#include https://github.com/abpvn/abpvn/blob/master/filter/abpvn_content_blocker.txt /exclude="../../exclusions.txt"
+!#else
+!#include https://raw.githubusercontent.com/abpvn/abpvn/master/filter/abpvn_adguard.txt /exclude="../../exclusions.txt"
+!#endif
--- a/filters/ThirdParty/filter_214_ABPVNList/template.txt
+++ b/filters/ThirdParty/filter_214_ABPVNList/template.txt
@@ -1,2 +1,2 @@
 !
-@include "https://raw.githubusercontent.com/abpvn/abpvn/master/filter/abpvn_adguard.txt" /exclude="../../exclusions.txt"
+@include platforms.txt


filters/ThirdParty/filter_214_ABPVNList/template.txt
```diff
@@ -1,2 +1,2 @@
 !
-@include "https://raw.githubusercontent.com/abpvn/abpvn/master/filter/abpvn_adguard.txt" /exclude="../../exclusions.txt"
+@include platforms.txt
```
ameshkov, line 2: `/exclude` should also be specified

ameshkov (changes requested):

filters/ThirdParty/filter_214_ABPVNList/platforms.txt
```diff
@@ -0,0 +1,5 @@
+!#if (adguard_ext_android_cb)
+!#include https://github.com/abpvn/abpvn/blob/master/filter/abpvn_content_blocker.txt /exclude="../../exclusions.txt"
+!#else
```
AdamWr, line 3: It's something new for me.
Is it a correct syntax?
I cannot find anything about `!#else` here - https://kb.adguard.com/en/general/how-to-create-your-own-ad-filters#hints_syntax

ameshkov: @zloyden please check it locally. In the list authors slack it was just pseudo-code by me, don't take it too seriously:)

zloyden: I had tried to build this locally, but unfortunately, there was no success.
For now I have used our `platformsIncluded` and `platformsExcluded`.

slavaleleka: wow, new filter — isn't is too much?

I believe it can be solved with negation
```
! version for AdGuard Content Blocker only
!#if (adguard_ext_android_cb)
...
!#endif

! for all other AdGuard products
!#if (!adguard_ext_android_cb)
...
!#endif
```

ameshkov (changes requested):
Content blocker does not load filters from the server, we should do it with id=214

filters/ThirdParty/filter_214_ABPVNList/platforms.txt
zloyden: After our discussion, it works fine (checked local build). Thank you:)

slavaleleka (approved):

filters/ThirdParty/filter_214_ABPVNList/template.txt
```diff
@@ -1,2 +1,9 @@
+! version for AdGuard Content Blocker only
+!#if (adguard_ext_android_cb)
+@include https://raw.githubusercontent.com/abpvn/abpvn/master/filter/abpvn_content_blocker.txt /exclude="../../exclusions.txt"
```
AdamWr, line 3: It seems that there are missed `""`.
I mean here:
```
@include https://raw.githubusercontent.com/abpvn/abpvn/master/filter/abpvn_content_blocker.txt
```
If I'm not wrong in other templates this link is usually between `""` like this:
```
@include "https://raw.githubusercontent.com/abpvn/abpvn/master/filter/abpvn_content_blocker.txt"
```
or is not needed?
The same below.

zloyden: The build was fine without them, so it's not necessary. But we have the quotes in other templates, so added them.

AdamWr (approved):

slavaleleka (approved):

ameshkov (approved):

ameshkov (approved):

PR #692, opened 2022-12-06
update compiler with updated scriptlets to v1.7.10 and extended-css to v2.0.24

stanislav-atr:

PR #699, opened 2022-12-14
update compiler to 1.1.68 with scriptlets v1.7.13, tsurlfilter v1.0.63

stanislav-atr:


AdamWr:
To be honest, I'm not sure what should be checked out there.
Compared to previous similar pull request, previously `@adguard/extended-css@` has been replaced by a new version - https://github.com/AdguardTeam/FiltersRegistry/pull/692/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deL5-R8
But now it looks like there will be two versions:
https://github.com/AdguardTeam/FiltersRegistry/blob/192d65805c41b7cee4ac473a1022808a9173e6c4/yarn.lock#L5-L13
Is it supposed to be like this?

slavaleleka:
> Is it supposed to be like this?

yes, it is absolutely normal

AdamWr (approved):

Alex-302 (approved):

PR #705, opened 2022-12-21
Exclused known invalid rules to improve readability of the log

Alex-302:


filters/exclusions.txt
```diff
@@ -250,18 +250,167 @@ _campaign=$popup
...
+~yandex.net,yandex.*#?#+js(set-constant, Object.prototype._getAdConfig, noopFunc)
+~yandex.net,yandex.*#?#+js(set-constant, Object.prototype._hasAdblock, noopFunc)
+~yandex.net,yandex.*#?#+js(set-constant, Object.prototype.getHoneypot, null)
+~yandex.net,yandex.*#?#+js(set-constant, Object.prototype.getYaads, null)
+~yandex.net,yandex.*#?#+js(set-constant, Object.prototype.initWidgetBlock, falseFunc)
+~yandex.net,yandex.*#?#+js(set-constant, Object.prototype.isOpenImageUnauthPopupEnabled, false)
+~yandex.net,yandex.*#?#+js(set-constant, Object.prototype.loaderVer, noopFunc)
+~yandex.net,yandex.*#?#+js(set-constant, Object.prototype.los, null)
+~yandex.net,yandex.*#?#+js(set-constant, Object.prototype.useAbdBundle, false)
+~yandex.net,yandex.*#?#+js(set-constant, home.export.adb, {})
```
zloyden, line 361: Not a small list of the rules. Are they problematic? Perhaps we should consider disabling this filter?

Alex-302: No. These rules has a different than filter's name purpose.
I don't know used `#?#+js()` instead of `##+js()`
They work in uBO, but no info about `#?#+js()` in their doc.

zloyden: Ok.

zloyden (approved):

filters/exclusions.txt
```diff
@@ -250,18 +250,167 @@ _campaign=$popup
...
+!
+! Filter name: Polish GDPR-Cookies Filters
+nike.com##body[data-overlay="cookieModalSimple"]:watch-attr(data-overlay):not([data-overlay="true"]) #gen-nav-commerce-header-v2 > div.pre-modal-window:has(> .pre-modal > .pre-modal-view > .pre-cookie-modal-body)
+!
+! Filter name: Polish Social Filters
+oszczedzarka.co.uk.textwidget > p > img[src="/img/ico/skype.png"]
+!
+! Filter name: road-block light
+science.hotnews.ro###\$\{tile\.name\}_\$\{tile\.id\}"
+||imasdk.googleapis.com/js/sdkloader/ima3.js$script,important,redirect=google-ima.js,domain=a1.ro|antena3.ro|antenastars.ro|as.ro|jurnalul.ro|observator.tv|observatornews.ro|spynews.ro|zutv.ro
```
AdamWr, line 382: If I'm not wrong `redirect=google-ima.js` is used by uBlock Origin, so maybe we should add an alias for it.

AdamWr (approved):

filters/exclusions.txt
Alex-302: https://github.com/AdguardTeam/FiltersCompiler/issues/167

jellizaveta (approved):

Sergey-Lyapin (approved):

Issue #709, opened 2022-12-26
Expand the list of pseudo-classes excluded for Android Content Blocker


slavaleleka:
### Prerequisites

- [X] This problem is not related to filtering issues. Otherwise, you should read this [article](https://kb.adguard.com/technical-support/reporting-tool);
- [X] You're using an up-to-date version of the problematic filter;

### Problem description

there are exclusions for few pseudo-classes for Android Content Blocker
https://github.com/AdguardTeam/FiltersRegistry/blob/master/custom_platforms.js#L310

### Proposed solution

add more supported extended pseudo-classes to the list
https://github.com/AdguardTeam/ExtendedCss#homepage

### Privacy

- [X] I agree to follow this condition

PR #710, opened 2022-12-27
Expand the list of pseudo-classes excluded for Android Content Blocker

zloyden:
https://github.com/AdguardTeam/FiltersRegistry/issues/709.
I'm not sure of the exclusion `:is()`, `:not()`.


custom_platforms.js
```diff
@@ -311,6 +311,14 @@ module.exports = {
...
                 ":contains\\(",
                 ":matches-css\\(",
+                ":matches-attr\\(",
+                ":matches-property\\(",
+                ":xpath\\(",
+                ":nth-ancestor\\(",
+                ":upward\\(",
+                ":remove\\(",
+                ":is\\(",
+                ":not\\(",
```
slavaleleka, line 321: both `:is()` and `:not()` can be standard pseudo-classes. should they be excluded as well?

zloyden: In fact, adding them was questionable and the only reason was that they are the part of the https://github.com/AdguardTeam/ExtendedCss. I wouldn't exclude them, what's your opinion?
I'll edit the description.

slavaleleka (approved):

PR #728, opened 2023-02-13
Replacing my lists' subscription URLs due to raw.githubusercontent.com problems

DandelionSprout:
For unknown reasons (I've received no info from GitHub about it), the sync time between my list edits and the time they show up on `raw.githubusercontent.com`, has recently increased from 5~9min, to somewhere more than 4 *hours*, whereas CDN services and GitLab CI/CD mirroring takes much less than 4 hours before listfixes go live.

It was a 50:50 call whether to use JSDelivr or GitLab, with neither site's ping limits being likely to be affected by Filters Registry's fetching of lists from them; I went with JSDelivr for 3 of them. For *Annoyances List* I used GitLab due to JSDelivr having possible incompatibilites regarding non-suffix files.

### DIFF
--- a/filters/ThirdParty/filter_249_NorwegianList/metadata.json
+++ b/filters/ThirdParty/filter_249_NorwegianList/metadata.json
@@ -7,7 +7,7 @@
   "expires": "4 days",
   "displayNumber": 249,
   "groupId": 7,
-  "subscriptionUrl": "https://raw.githubusercontent.com/DandelionSprout/adfilt/master/NorwegianExperimentalList%20alternate%20versions/NordicFiltersAdGuard.txt",
+  "subscriptionUrl": "https://cdn.jsdelivr.net/gh/DandelionSprout/adfilt@master/NorwegianExperimentalList%20alternate%20versions/NordicFiltersAdGuard.txt",
   "tags": [
     "recommended",
     "purpose:ads",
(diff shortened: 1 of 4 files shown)


zloyden (changes requested):
@DandelionSprout please also exchange your subscription URLs in templates.

DandelionSprout:
I'll look into that within 18 hours.

DandelionSprout:
Having woke up and been made aware of https://github.com/orgs/community/discussions/46691, I'll keep you guys updated on whether `raw.githubusercontent.com` becomes viable again soon or not.

jimaek:
Let me know if jsDelivr can help with anything :)

DandelionSprout:
>Let me know if jsDelivr can help with anything :)

Now that you mention it, it seems there's a minor problem with jsDelivr where it tries to download files with file extensions to the local computer instead of displaying them online, e.g. https://cdn.jsdelivr.net/gh/DandelionSprout/adfilt@master/AnnoyancesList

(Adblockers work in such way that I unfortunately can't simply change the filename to `AnnoyancesList.txt` without heavy coordination with adblocker devs and end-users.)

MartinKolarik:
@DandelionSprout the downloading happens because without the extension we have no way of knowing the correct mime type of the file so we set a generic one, which then Chrome doesn't know how to display. Anyway, it doesn't change anything unless you expect the files to be directly opened by end users. When fetched via JavaScript, those files will work the same way as the ones with extensions.

jimaek:
If the AdGuard team is interested in a deeper integration we could also setup a custom proxy config for all lists, similar to what we did with Ghost https://www.jsdelivr.com/oss-cdn/ghost

ameshkov:
@jimaek hi Dmitriy, thank you, but we serve filter lists from our own domain. I guess this may be helpful to other filters maintainers like @DandelionSprout though.

PR #729, opened 2023-02-17
exclude ##.adsbygoogle

Yuki2718:
For https://github.com/easylist/easylist/commit/c09ae084dc666529d0ae0e639836f423bd1231ff

BTW isn't `^` needed for `/###banner_ad$/` as well? Specific `###banner_ad` may be used by good reason.

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -36,10 +36,10 @@ $popup,~third-party
 !
 !### Anti-adblock baits ###
 ! all filters
-/^##\.adsbygoogle$/
-/###banner_ad$/
-/^##\.adsbox$/
-/^##\.afs_ads$/
+/^(~.*)?##\.adsbygoogle$/
+/^(~.*)?###banner_ad$/
+/^(~.*)?##\.adsbox$/
+/^(~.*)?##\.afs_ads$/
 !
 !### Easylist ###
 !
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -36,6 +36,7 @@ $popup,~third-party
 !
 !### Anti-adblock baits ###
 ! all filters
+/^##\.adsbygoogle$/
 /###banner_ad$/
 /^##\.adsbox$/
 /^##\.afs_ads$/


AdamWr:
I guess it's `/###banner_ad$/`, because there is a rule like this `~bg-ikonomika.com###banner_ad` in the `Bulgarian list`.
Maybe it should be changed to something like - `/^(~.*)?###banner_ad$/`?

By the way, now I have noticed that also `ROList` has rule like this `~4tuning.ro##.adsbygoogle` and `xinggsf` has `~google.com##.adsbygoogle`

So maybe it also should be - `/^(~.*)?##\.adsbygoogle$/`?

AdamWr (approved):

zloyden (approved):

piquark6046 (approved):

Yuki2718:
Now I think of, there can theoretically be this type of specific filter: `~a.foo.com,foo.com##.adsbygoogle`
So I suggest to change to `/^(~[-.a-z]+,?)*##\.adsbygoogle$/`

AdamWr:
That's a good point, but if I'm not wrong, it will not cover all cases.
For example, it will not match domains with digits or with "punycode".
Maybe something like this:
```
^(~([-.\w\d*]|[^\x00-\x7F])+,?)*##\.adsbygoogle$
```
will be okay (https://regex101.com/r/LLqMr0/1), but I'm still not sure if all cases will be catched.

filters/exclusions.txt
```diff
@@ -36,9 +36,10 @@ $popup,~third-party
 !
 !### Anti-adblock baits ###
 ! all filters
-/###banner_ad$/
-/^##\.adsbox$/
-/^##\.afs_ads$/
+/^(~.*)?##\.adsbygoogle$/
+/^(~.*)?###banner_ad$/
+/^(~.*)?##\.adsbox$/
+/^(~.*)?##\.afs_ads$/
```
Alex-302, line 42: why changed?

Yuki2718: Can there be harm? Some filter may add `~foo.com##.afs_ads` in future.

Alex-302: Ok.

Issue #756, opened 2023-03-14
Exclude '$removeparam', '$removeheader', '$extension', and '$csp' rules for not supported platforms


slavaleleka:
### Prerequisites

- [X] This problem is not related to filtering issues. Otherwise, you should read this [article](https://kb.adguard.com/technical-support/reporting-tool);
- [ ] You're using an up-to-date version of the problematic filter;

### Problem description

Modifiers `$removeparam`, `$removeheader`, `$extension`, and `$csp` are definitely not supported by AdGuard for iOS and AdGuard for Safari:
https://adguard.com/kb/general/ad-filtering/create-own-filters/#removeparam-modifier-compatibility
https://adguard.com/kb/general/ad-filtering/create-own-filters/#removeheader-modifier
https://adguard.com/kb/general/ad-filtering/create-own-filters/#extension-modifier
https://adguard.com/kb/general/ad-filtering/create-own-filters/#csp-modifier

But there are such rules in built filters lists. To make sure, search for `$removeparam=`, `$removeheader=`, and `$csp` in these filters:
https://filters.adtidy.org/ios/filters/19_optimized.txt
https://filters.adtidy.org/extension/safari/filters/21.txt
https://filters.adtidy.org/ios/filters/2_optimized.txt

### Proposed solution

`custom_platforms.js` should be modified as such rules should be excluded during the build. check these modifiers for  other platforms as well

### Privacy

- [X] I agree to follow this condition

PR #757, opened 2023-03-15
Remove some unsupported modifiers for iOS, Safari

zloyden:
https://github.com/AdguardTeam/FiltersRegistry/issues/756
~~[Regex](https://regex101.com/r/MhdbBh/1)~~ [replaced by simpler ones](https://github.com/AdguardTeam/FiltersRegistry/pull/757#discussion_r1136906671).

### DIFF
--- a/custom_platforms.js
+++ b/custom_platforms.js
@@ -87,7 +87,10 @@ module.exports = {
             "removeRulePatterns": [
                 "\\$extension",
                 ",extension",
-                "\\$(.*,)?remove(param|header)",
+                "$removeparam",
+                ",removeparam",
+                "$removeheader",
+                ",removeheader",
                 "^((?!#%#).)*\\$\\$|\\$\\@\\$",
                 "\\$(.*,)?mp4",
                 "\\$(.*,)?replace=",
@@ -185,7 +188,10 @@ module.exports = {
         "path": "extension/safari",
         "configuration": {
             "removeRulePatterns": [
-                "\\$(.*,)?remove(param|header)",
+                "$removeparam",
+                ",removeparam",
+                "$removeheader",
+                ",removeheader",
                 "^((?!#%#).)*\\$\\$|\\$\\@\\$",
                 "\\$(.*,)?mp4",
                 "\\$(.*,)?replace=",
--- a/custom_platforms.js
+++ b/custom_platforms.js
@@ -85,6 +85,9 @@ module.exports = {
         "path": "ios",
         "configuration": {
             "removeRulePatterns": [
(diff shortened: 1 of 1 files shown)


custom_platforms.js
```diff
@@ -85,6 +85,9 @@ module.exports = {
         "path": "ios",
         "configuration": {
             "removeRulePatterns": [
+                "\\$extension",
+                ",extension",
+                "\\$remove(param|header)",
```
AdamWr, line 90: `removeparam` and `removeheader` can be also used with other modifiers, for example:
```adblock
||example.org^$xmlhttprequest,removeparam=param
||example.org^$xmlhttprequest,removeheader=location
```
so I guess that these kind of rules should be also excluded.

zloyden: In fact, I only see the rules like below, but maybe it will be useful for future ones.
```adblock
||example.org^$removeheader=location
```
@AdamWr what about this? Seems it removes the rules you mentioned fine.
```
(\\$|,)remove(param|header)
```

zloyden: As I can see on those files, there is no difference, but the pattern above seems a bit specific, so I've replaced it.

AdamWr (approved):

custom_platforms.js
```diff
@@ -182,6 +185,7 @@ module.exports = {
         "path": "extension/safari",
         "configuration": {
             "removeRulePatterns": [
+                "\\$(.*,)?remove(param|header)",
```
slavaleleka, line 188: it would be great to have here examples of rules for regexp, isn't it?
so it could be easier to fix issues if any occur

for example:
```
"removeRulePatterns": [
    // ||example.org^$removeheader=location
    // ||example.org^$xmlhttprequest,removeparam=param
    // ||example.org^$xmlhttprequest,removeheader=location
    "\\$(.*,)?remove(param|header)",
```

zloyden: I'm not sure if it's necessary to do that, the regex looks fine.
I've updated the description with a link to examples from `regex101.com`.

custom_platforms.js
```diff
@@ -85,6 +85,9 @@ module.exports = {
         "path": "ios",
         "configuration": {
             "removeRulePatterns": [
+                "\\$extension",
+                ",extension",
+                "\\$(.*,)?remove(param|header)",
```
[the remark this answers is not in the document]
zloyden: Done.

Alex-302 (approved):

AdamWr (approved):

Issue #786, opened 2023-05-05
Add EasyList Portuguese Filter


DandelionSprout:
### Prerequisites

- [X] I checked the documentation and found no answer;
- [X] I checked to make sure that this issue has not already been filed;
- [X] This is not an ad/bug report.

### Problem description

I've noticed that a fair few entries in [https://easylist-downloads.adblockplus.org/easylistportuguese.txt](https://easylist-downloads.adblockplus.org/easylistportuguese.txt) are not present in AdGuard Spanish/Portuguese Filter, especially ones for the `.cv`, `.ao`, and `.mz` TLDs.

Since most other AdGuard regional lists tend to incorporate ABP-affiliated regional lists, could the same be done for EasyList Portuguese?

### Proposed solution

Implement entries from EasyList Portuguese into AdGuard Spanish/Portuguese Filter.

### Additional information

EasyList Portuguese is a relatively new list, created in what I think was 2021, long after AdGuard Spanish/Portuguese Filter was created. So it's understandable that this implementation hasn't been done yet.

Additionally, EasyList Portuguese is (to the best of my knowledge) the only ABP-affiliated regional list to not be included at all in uBO nor AdGuard products, making it additionally important to make use of its entries somehow.

zloyden:
I think that we can add it as a separated filter.

PR #789, opened 2023-05-10
Adding new EasyList filter - EasyList Portuguese

zloyden:
https://github.com/AdguardTeam/FiltersRegistry/issues/786


aalpatkina (approved):

filters/ThirdParty/filter_255_EasyListPortuguese/metadata.json
```diff
@@ -0,0 +1,18 @@
+{
+  "filterId": 255,
```
Alex-302, line 2: For EasyList filters used 100-199, so replace all 255.

AdamWr (approved):

filters/ThirdParty/filter_255_EasyListPortuguese/metadata.json
```diff
@@ -0,0 +1,18 @@
...
+  "name": "EasyList Portuguese",
+  "description": "Additional filter list for websites in Spanish, Portuguese.",
```
Alex-302, line 4: `Spanish, Portuguese` maybe better `and` instead of `,`? (also in another files)

piquark6046 (approved):

Alex-302 (changes requested):

filters/ThirdParty/filter_255_EasyListPortuguese/metadata.json
zloyden: Done.

filters/ThirdParty/filter_255_EasyListPortuguese/metadata.json
zloyden: I guess there is no difference in the meaning, but there is only 2 languages, so we can use `and` instead of the comma.

aalpatkina (approved):

aalpatkina (approved):

PR #793, opened 2023-05-17
Excluded bad rule in ADgk Mobile China list

Alex-302:
https://github.com/AdguardTeam/AdguardFilters/issues/149677#issuecomment-1546881388 https://github.com/banbendalao/ADgk/issues/74

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -42,7 +42,7 @@ $popup,~third-party
 /^(~.*)?##\.afs_ads$/
 !
 !### ADgk Mobile China list ###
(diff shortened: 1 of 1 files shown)


AdamWr (approved):

zloyden (approved):
I partially agree with this, but shouldn't it be fixed directly here https://github.com/banbendalao/ADgk/issues/74?

Alex-302:
> I partially agree with this, but shouldn't it be fixed directly here [banbendalao/ADgk#74](https://github.com/banbendalao/ADgk/issues/74)?

There is no activity last time.

Issue #800, opened 2023-06-11
Remove AdGuard DNS filter from the FiltersRegistry


ameshkov:
### Prerequisites

- [X] I checked the documentation and understood it;
- [X] I checked to make sure that this issue has not already been filed;

### Problem description

We now have a separate HostlistsRegistry and all AdGuard products have their own DNS filtering modules and blocklists.

There's no need to keep AdGuard DNS filter here anymore.

### Proposed solution

I suggest marking it as obsolete so that the programs stopped showing it on the filters screen and removed it from the filters meta.

Before doing it, we'll need to check if this change can cause any issues with the old versions so don't merge until the check is done.

slavaleleka:
few kb articles should be updated after it:
- https://github.com/AdguardTeam/KnowledgeBase/blob/master/docs/general/ad-filtering/adguard-filters.md?plain=1#L38
- https://github.com/AdguardTeam/KnowledgeBase/blob/master/docs/adguard-for-ios/solving-problems/system-wide-filtering.md?plain=1#L45

PR #801, opened 2023-06-13
Remove AdGuard DNS filter from the FiltersRegistry

zloyden:
https://github.com/AdguardTeam/FiltersRegistry/issues/800

### DIFF
--- a/filters/filter_15_DnsFilter/metadata.json
+++ b/filters/filter_15_DnsFilter/metadata.json
@@ -1,6 +1,6 @@
 {
   "filterId": 15,
-  "name": "AdGuard DNS filter",
+  "name": "(Obsolete) AdGuard DNS filter",
   "description": "Filter composed of several other filters (AdGuard Base filter, Social Media filter, Tracking Protection filter, Mobile Ads filter, EasyList and EasyPrivacy) and simplified specifically to be better compatible with DNS-level ad blocking.",
   "timeAdded": 1404115015843,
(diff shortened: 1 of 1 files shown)


filters/filter_15_DnsFilter/metadata.json
```diff
@@ -20,7 +20,9 @@
     "reference:118",
     "reference:102",
     "reference:13",
-    "reference:16"
+    "reference:16",
+    "obsolete"
   ],
-  "trustLevel": "full"
+  "trustLevel": "full",
+  "disabled": true
```
ameshkov, line 27: Not sure we should make it `disabled` though, let it continue updating for some time.

filters/filter_15_DnsFilter/metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "filterId": 15,
-  "name": "AdGuard DNS filter",
+  "name": "(Obsolete) AdGuard DNS filter",
```
ameshkov, line 3: The list itself is not obsolete so I suggest not changing the name

zloyden: Done.

filters/filter_15_DnsFilter/metadata.json
zloyden: So, we can add it later?
I left only the tag "obsolete" as you suggested earlier.

Alex-302 (approved):

PR #818, opened 2023-07-31
Update compiler to v1.1.93

maximtop:
Update should fix this bug
https://github.com/AdguardTeam/FiltersCompiler/issues/185

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,9 +2,9 @@
   "name": "adguard-filters",
   "version": "1.0.53",
   "description": "AdGuard filters registry",
-  "homepage": "http://adguard.com",
+  "homepage": "https://adguard.com",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


slavaleleka (approved):

Alex-302:
Wit this version I see the same rule, like in @Yuki2718 comment. 
Which rule is the expected result?

maximtop:
> Wit this version I see the same rule, like in @Yuki2718 comment. Which rule is the expected result?

have you run yarn install before checking?

Alex-302:
> have you run yarn install before checking?

sure

Alex-302:
After `yarn install --force` it works.
Now resulting rule is
`webhostingpost.com##+js(nano-setInterval-booster, \,dataType:_, 1000, 0.02)`

PR #823, opened 2023-08-25
update compiler to v1.1.96

slavaleleka:
as usual run build and check the report.txt please — previously invalid scriptlet rules should be valid now

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "adguard-filters",
-  "version": "1.0.54",
+  "version": "1.0.55",
   "description": "AdGuard filters registry",
   "homepage": "https://adguard.com",
   "dependencies": {
-    "adguard-filters-compiler": "git+https://github.com/AdguardTeam/FiltersCompiler.git#v1.1.94"
+    "adguard-filters-compiler": "git+https://github.com/AdguardTeam/FiltersCompiler.git#v1.1.96"
   },
   "main": "index.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "AdGuard filters registry",
   "homepage": "https://adguard.com",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


AdamWr:
>! Rule "jeger.pl##+js(set-cookie, _acceptCookies, 1, , reload, 1)" converted to: "jeger.pl#%#//scriptlet('ubo-set-cookie.js', '_acceptCookies', '1', '', 'reload', '1')"

I guess that rules with `reload` as a 4th argument should be converted to `set-cookie-reload` scriptlet.
https://github.com/gorhill/uBlock/wiki/Resources-Library#set-cookiejs-

slavaleleka:
> `set-cookie` rules with `reload` should be converted to `set-cookie-reload`

there is another task for this:
https://github.com/AdguardTeam/FiltersCompiler/issues/181

AdamWr:
Okay, but now it will be converted to "incorrect" rule which I guess will not work how it should.

Another problem is with `offsetExpiresSec` argument, for example:
>! Rule "amway.pl##+js(set-cookie, isAnalyticsCookiesAccepted, false, 1year)" converted to: "amway.pl#%#//scriptlet('ubo-set-cookie.js', 'isAnalyticsCookiesAccepted', 'false', '1year')"

As far as I can see, `offsetExpiresSec` argument is supported only in `trusted-set-cookie`/`trusted-set-cookie-reload` - https://github.com/AdguardTeam/Scriptlets/blob/master/wiki/about-trusted-scriptlets.md#-%EF%B8%8F-trusted-set-cookie
but not in https://github.com/AdguardTeam/Scriptlets/blob/master/wiki/about-scriptlets.md#set-cookie

In `set-cookie` it will be set as a `path`, so it will be probably just ignored.

slavaleleka:
> `#%#//scriptlet('trusted-click-element', ...` are not converted

trusted scriptlets are not supposed to be converted by design

AdamWr:
Okay, so if these issues are planned to be fixed separately, then it looks good to me.

AdamWr (approved):

zloyden (approved):

PR #825, opened 2023-09-05
fler.cz - Remove f_gid from blocked cookies

RikudouSage:
The cookie is needed for logging in.


Alex-302:
Wrong repository. Removed.

PR #831, opened 2023-10-16
Upd optimization cfg for all AdGuard Annoyances filters + Chinese

Alex-302:


optimization_config.json
```diff
@@ -25,6 +25,12 @@
...
   },
+  {
+    "filterId": 223,
+    "percent": 50,
+    "strict": true,
+    "comment": "AdGuard Chinese filter"
+  },
   {
     "filterId": 14,
     "percent": 60,
```
slavaleleka, line 36: main annoyances filter must be large. what about decreasing its percent?

ameshkov: it's actually not that bad

Alex-302: main annoyances filter must be removed:)

ameshkov (changes requested):

slavaleleka (approved):

slavaleleka (approved):

PR #835, opened 2023-10-23
Exclude YouTube scriptlet rule in 3rd-party filters

### LINKED ISSUE #834: Exclude YouTube scriptlet rule in 3rd-party filters
### Prerequisites

- [ ] You're using an up-to-date version of the problematic filter;

### Problem description

Because of the recent anti-adblock war of YouTube, AdGuard and uBlock Origin receive many related issues.

In xinggsf filter that is registered in [AdguardTeam/FiltersRegistry](https://github.com/AdguardTeam/FiltersRegistry), there is a problematic rule triggering the anti-adblock alert. 
https://github.com/uBlockOrigin/uAssets/issues/19976#issuecomment-1774090029

### Proposed solution

To reduce the issue and prevent a similar rule, a regexp can be added into `exclusions.txt` file.
A pull request will be opened.

### Privacy

piquark6046:
Fix #834

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -236,7 +236,7 @@ _campaign=$popup
 !###########################
 !
 !### xinggsf ###
-/(?<!m\.)(www\.)?youtube\.com(,)?.*#%#\/\/scriptlet\(/
+#%#//scriptlet('ubo-json-prune.js', '[].playerResponse.adPlacements [].playerResponse.playerAds playerResponse.adPlacements playerResponse.playerAds adPlacements playerAds')
 /##\[id\|=ad\]$/
 /##\.adsbygoogle,\.ads$/
 */market-$script


filters/exclusions.txt
```diff
@@ -236,6 +236,7 @@ _campaign=$popup
 !###########################
 !
 !### xinggsf ###
+/(?<!m\.)(www\.)?youtube\.com(,)?.*#%#\/\/scriptlet\(/
```
AdamWr, line 239: As far as I understand, it will disable all scriptlet rules for youtube from third party filter lists.
Currently I see scriptlets for youtube only in `xinggsf` filter, so it's not a problem, but in the future there are might be useful scriptlets in other lists as well.
So, maybe it would be better to exclude only these which are problematic?

piquark6046:
```suggestion
(?<!m\.)(www\.)?youtube\.com(,)?.*#%#\/\/scriptlet\(['"]{1}(ubo-)?json-prune[\.a-z'"]+, *['"]{1}(\[\]\.playerResponse\.adPlacements|\[\]\.playerResponse\.playerAds|playerResponse\.adPlacements|playerResponse\.playerAds|adPlacements\.playerAds)
```
https://regex101.com/r/3bBQWc/1

Like this? Or, just a literally the rule?

Alex-302: original rule or a part can be added as is without `/regex/`

piquark6046:
```suggestion
#%#//scriptlet('ubo-json-prune.js', '[].playerResponse.adPlacements [].playerResponse.playerAds playerResponse.adPlacements playerResponse.playerAds adPlacements playerAds')
```
@Alex-302 Like this?

AdamWr: If only a specific rule causes it, then I think that we can exclude one specific rule.
If there will be more rules causing detection, then we can use regexp.
But it's just my opinion, so maybe better to wait and see what other @AdguardTeam/filters-maintainers say.

piquark6046: It is owned by @xinggsf.

iam-py-test: > Btw, can the author do this on his own?

A PR has been submitted but not yet merged: https://github.com/xinggsf/Adblock-Plus-Rule/pull/86
Thanks

zloyden (approved):

PR #850, opened 2023-11-21
exclude unsupported modifiers for ext_safari, ios, and ext_android_cb platforms

slavaleleka:
https://github.com/AdguardTeam/FiltersCompiler/issues/201


Alex-302 (approved):

custom_platforms.js
```diff
@@ -223,6 +223,7 @@ module.exports = {
                 ",empty",
                 "\\$webrtc",
                 "\\$csp",
+                "\\$content$",
```
AdamWr, line 226: I'm just curious if there is any reason to use `\\$content$` instead of `\\$content`.
If so, maybe it should be updated for other platforms?

By the way, shouldn't something like this `,content(,|$)` or just `,content` be also added?
In the case if there would be rules like these:
```adblock
@@||example.com^$elemhide,content
@@||example.org^$elemhide,content,xmlhttprequest
```

Alex-302: @AdamWr `content` is ignored in Content Blockers:) As I remember, all these rules required for apps only, when something is broken by AdGuard's injections.

Alex-302: https://github.com/AdguardTeam/AdguardFilters/commit/05ca3dc47a25e5479e30bbad69b5c77c4be2e8a2

AdamWr: If these rules are required only for CoreLibs apps, then I think that should be excluded from filter lists for content blockers by default (the same as it's done for Chromium extensions).

AdamWr: By the way, if I'm not wrong, current solution (`",content"`) for Chromium based browsers may not works correctly.
It excludes all rules with `,content`, so rule like this:
```adblock
giga.de#@%#(function(o){function a(a){return{get:function(){return a},set:b}}function b(){}function c(){throw"Adguard: stopped a script execution.";}var d={},e=a(function(a){a(!1)}),f={},g=EventTarget.prototype.addEventListener;o(d,{spid_control_callback:a(b),content_control_callback:a(b),vid_control_callback:a(b)});o(f,{config:a(d),_setSpKey:{get:c,set:c},checkState:e,isAdBlocking:e,getSafeUri:a(function(a){return a}),pageChange:a(b),setupSmartBeacons:a(b)});Object.defineProperty(window,"_sp_",a(f));EventTarget.prototype.addEventListener=function(a){"sp.blocking"!=a&&"sp.not_blocking"!=a&&g.apply(this,arguments)}})(Object.defineProperties);
```
is excluded from https://filters.adtidy.org/extension/chromium/filters/14.txt but it's in https://filters.adtidy.org/android/filters/14.txt

slavaleleka: >  if there is any reason to use \\$content$ instead of \\$content

yes, to avoid exclusions of such rules:
```
@@||example.com^$elemhide,content
@@||example.org^$elemhide,content,xmlhttprequest
```
where `$elemhide`, for example, still works in AdGuard for Safari

slavaleleka: > current solution (",content") for Chromium based browsers may not works correctly

could you please create a PR with fix for `custom_platforms.js`?

AdamWr: Sure, I can try to do it.
By the way, should it be done on BitBucket, here or in both places?

slavaleleka: here is enough
when the FiltersRegistry PR is merged, I'll copy the changes to FiltersCompiler

AdamWr: Okay

PR #853, opened 2023-11-22
Fix #852 Improve removeRulePatterns

### LINKED ISSUE #852: Improve removeRulePatterns
### Prerequisites

- [X] You're using an up-to-date version of the problematic filter;

### Problem description

All modifiers from `removeRulePatterns` should be excluded only if occurs after (latest?) `$` character.

For example, if there is exclusion like this:
```
,content
```
or
```
,cookie
```
then all rules containing this will be exclued.

For example:
```adblock
giga.de#@%#(function(o){function a(a){return{get:function(){return a},set:b}}function b(){}function c(){throw"Adguard: stopped a script execution.";}var d={},e=a(function(a){a(!1)}),f={},g=EventTarget.prototype.addEventListener;o(d,{spid_control_callback:a(b),content_control_callback:a(b),vid_control_callback:a(b)});o(f,{config:a(d),_setSpKey:{get:c,set:c},checkState:e,isAdBlocking:e,getSafeUri:a(function(a){return a}),pageChange:a(b),setupSmartBeacons:a(b)});Object.defineProperty(window,"_sp_",a(f));EventTarget.prototype.addEventListener=function(a){"sp.blocking"!=a&&"sp.not_blocking"!=a&&g.apply(this,arguments)}})(Object.defineProperties);
```
is excluded from https://filters.adtidy.org/extension/chromium/filters/14.txt
But it's in https://filters.adtidy.org/android/filters/14.txt

The same for these rules:
```adblock
bridalgown.work,contents-group.work,heisei-housewarming.work,liquidfoundation.work,nailcolor.work,studioglass.work,tapestry.work,teaceremony.work,weddinghall.work#%#//scriptlet("set-constant", "blockId", "0")
bridalgown.work,contents-group.work,heisei-housewarming.work,liquidfoundation.work,nailcolor.work,studioglass.work,tapestry.work,teaceremony.work,weddinghall.work###pr_ad
```
These rules are excluded from https://filters.adtidy.org/extension/chromium/filters/7.txt
But are in https://filters.adtidy.org/windows/filters/7.txt

Probably there may be also special cases like `$path` or `$domain` which should be taken into consideration to not exclude rule like:
```adblock
[$path=page.html]##.textad,cookie
```

### Proposed solution

Improve regexp.

### Privacy

AdamWr:
Fixes https://github.com/AdguardTeam/FiltersRegistry/issues/852

Two examples:
https://regex101.com/r/80qB2f/1
https://regex101.com/r/SuPEko/1


custom_platforms.js
```diff
@@ -68,20 +68,16 @@ module.exports = {
                 "^((?!#%#).)*\\$\\$|\\$\\@\\$",
                 "\\$(.*,)?replace=",
                 "important,replace=",
-                "\\$(.*,)?app",
+                "\\$(?!((path|domain)=.*])).*app=",
```
slavaleleka, line 71: maybe it's time to make separate constants for such patterns so they can be reused and properly described?

AdamWr: Or maybe we could have two separated things, like `removeRulePatterns` and `removeRuleModifiers`?
So, in `removeRuleModifiers` we could simply just add modifiers without having to use regular expression.

slavaleleka: sounds like a separate task for FiltersCompiler

and it doesn't look easy now because FiltersCompiler simply matches the regexp pattern for `removeRulePatterns`, but _rules parsing_ is required for `removeRuleModifiers` (if I understand the feature request correctly)

slavaleleka: but it may be implemented in the future still. if you're sure it may be useful, file an issue please

AdamWr: I just thought that it could be simply done like this:
<details><summary>Code:</summary>

It's proof of concept and might not cover all cases.

```js
const removeRuleModifiers = ['extension', 'app', 'content'];

const modifierDelimeter = '$';
const pathModifier = '[$path=';
const domainModifier = '[$domain=';
const checkIfModifierSupported = rule => {
  if (!rule.includes(modifierDelimeter) || rule.includes(pathModifier) || rule.includes(domainModifier)) {
    // do nothing if there isn't "$" or it's [$path=] or [$domain=] modifier
    return;
  }
  const delimeterIndex = rule.lastIndexOf(modifierDelimeter);
  const modifiers = rule.substring(delimeterIndex + 1).split(',');
  if (modifiers.some(el => removeRuleModifiers.includes(el))) {
    return false;
  }
  return true;
};

checkIfModifierSupported('@@||test.com^$content,elemhide'); // false
checkIfModifierSupported('ads/blabla$domain=examle.com,elemhide,content,xmlhttprequest'); // false
checkIfModifierSupported('@@||test.com^$elemhide'); // true
checkIfModifierSupported('ads/blabla$domain=content.com,elemhide,xmlhttprequest'); // true
```

</details>

But if you say that it's not easy then fine.

<hr>

>maybe it's time to make separate constants for such patterns so they can be reused and properly described?

Could you please provide an example how and where this constant should be added?

slavaleleka: before `module.exports` in the custom_platforms.js:

```
/**
 * <description>
 *
 * @example
 * <examples>
 */
const REPLACE_MODIFIER_PATTERNS = [
    "\\$(.*,)?replace=",
    "important,replace=",
];

module.exports = { ... }
```

and then just 
```
...REPLACE_MODIFIER_PATTERNS,
```

Alex-302 (approved):

zloyden (approved):

AdamWr:
As suggested in https://github.com/AdguardTeam/FiltersRegistry/pull/853#discussion_r1402413911, I will add constants and description for each pattern - it's still in progress.
I will inform when it's ready for review.

AdamWr:
Done, it should be ready for review.

custom_platforms.js
```diff
@@ -63,25 +516,16 @@ module.exports = {
...
+                ...HTML_FILTERING_MODIFIER_PATTERNS,
+                ...REPLACE_MODIFIER_PATTERNS,
+                ...APP_MODIFIER_PATTERNS,
+                ...NETWORK_MODIFIER_PATTERNS,
+                ...PROTOBUF_MODIFIER_PATTERNS,
+                ...EXTENSION_MODIFIER_PATTERNS,
+                ...HLS_MODIFIER_PATTERNS,
+                ...JSONPRUNE_MODIFIER_PATTERNS,
+                ...REFERRERPOLICY_MODIFIER_PATTERNS,
+                ...CONTENT_MODIFIER_PATTERNS,
```
slavaleleka, line 528: list of removeRulePatterns for `EXTENSION_CHROMIUM`, `EXTENSION_EDGE` and `EXTENSION_OPERA` are same. lets use a single constant for this list
```
const CHROMIUM_BASED_EXTENSION_PATTERNS = [ ... ];
```

custom_platforms.js
```diff
@@ -6,6 +6,459 @@
...
+ * Pattern parts:
+ * - `\\$` — modifiers divider
+ * - `(?!((path|domain)=.*])).*` — negative lookahead to exclude rules like `[$path=...]##.textad,jsinject`
+ * - `.*jsinject` — jsinject modifier itself
+ * - `(,|$)` — end of line or modifiers divider, as `$jsinject` can be followed by other modifiers (`,`).
+ *
+ * @example
+ * ```@@://www.atlassian.com^$elemhide,jsinject,extension```
+ */
+const JSINJECT_MODIFIER_PATTERNS = [
```
slavaleleka, line 178: may it be added for other extensions, not just ext_ublock?

slavaleleka: the same for `URLBLOCK_MODIFIER_PATTERNS`

AdamWr: Yes, both can be added for `android-content-blocker`.

AdamWr: Done.

custom_platforms.js
AdamWr: Done.

custom_platforms.js
```diff
@@ -6,6 +6,496 @@
...
+    ...NETWORK_MODIFIER_PATTERNS,
+    ...PROTOBUF_MODIFIER_PATTERNS,
+    ...EXTENSION_MODIFIER_PATTERNS,
+    ...HLS_MODIFIER_PATTERNS,
+    ...JSONPRUNE_MODIFIER_PATTERNS,
+    ...REFERRERPOLICY_MODIFIER_PATTERNS,
+    ...CONTENT_MODIFIER_PATTERNS,
+];
+
+const SAFARI_BASED_EXTENSION_PATTERNS = [
```
slavaleleka, line 475:
```suggestion
/**
 * Used for `EXTENSION_SAFARI` and `IOS` platforms.
 */
const SAFARI_BASED_EXTENSION_PATTERNS = [
```

custom_platforms.js
```diff
@@ -6,6 +6,496 @@
...
+    ":matches-css-after\\(",
+    ":-abp-has\\(",
+    ":-abp-contains\\(",
+    "#\\?#",
+    "#\\$\\?#",
+    "#@\\?#",
+    "#@\\$\\?#",
+];
+
+const CHROMIUM_BASED_EXTENSION_PATTERNS = [
```
slavaleleka, line 462:
```suggestion
/**
 * Used for `EXTENSION_CHROMIUM`, `EXTENSION_EDGE`, and `EXTENSION_OPERA` platforms.
 */
const CHROMIUM_BASED_EXTENSION_PATTERNS = [
```

slavaleleka (approved):
finally it is refactored! great work :+1:

zloyden (approved):

custom_platforms.js
```diff
@@ -164,21 +600,14 @@ module.exports = {
...
-                "important,protobuf",
-                "\\$extension",
-                ",extension",
-                "\\$hls=",
-                ",hls=",
-                "\\$jsonprune=",
-                ",jsonprune=",
-                "\\$referrerpolicy",
-                ",referrerpolicy"
+                ...HTML_FILTERING_MODIFIER_PATTERNS,
```
AdamWr, line 603: By the way, do we still need to exclude HTML filtering rules in Firefox extension or perhaps it can be removed now?

Issue #866, opened 2023-12-10
Do not include *_specific_uBO.txt of Fanboy Annoyances/EasyList Cookie for Content blocker


Yuki2718:
### Prerequisites

- [X] I checked the documentation and understood it;
- [X] I checked to make sure that this issue has not already been filed;

### Problem description

https://github.com/AdguardTeam/AdguardFilters/issues/168084#issuecomment-1848875626
`*_specific_uBO.txt` assumes the platform the list is used can handle all advanced syntax, which is not the case for CB. This causes issues like the above.

### Proposed solution

Create another registry of Fanboy Annonyances/EasyList Cookie without `*_specific_uBO.txt` subcomponent for CB?

### Additional information

Or else convert all of their hiding rules to CSS rules with `display: none`.

krystian3w:
Also free iOS seems is CB (CSS and scriptlets may need buy licence).

If you have access to Rayan hosting (Web FTP, (s)FTP(s) or SSH) try:


```adb
!#if (!adguard_ext_android_cb && !adguard_app_ios)
!#include uBO_supplement.txt
!#endif
```

or

```adb
!#if !adguard_ext_android_cb
!#if !adguard_app_ios
!#include uBO_supplement.txt
!#endif
!#endif
```

Also possible use files created by gorhill (something like mirror).

krystian3w:
Second problem is rejection add many exceptions (10-20 or more broken pages) for ABP where Rayan repaired scrolling by CSS/scriptlet and skipped bug existed with ABP so also with CB (Android/iOS).

Yuki2718:
Now EP has `_specific_uBO.txt` too https://github.com/easylist/easylist/pull/18170/ and this will eventually be extended to EL.

PR #868, opened 2023-12-11
Increase trust level of Fanboy's Annoyance filters list

### LINKED ISSUE #867: Increase trust level of Fanboy's Annoyance filters list
### Prerequisites

- [X] You're using an up-to-date version of the problematic filter;

### Problem description

Currently, the configuration for Fanboy's Annoyance filters list uses uBO version of the filters list.
But, the value of its trust level is `low`.

As a result, many filters using advanced filtering syntax are deleted.

Related issue: https://github.com/AdguardTeam/AdguardFilters/issues/168277

### Proposed solution

Increase trust level of Fanboy's Annoyance filters list.

### Privacy

piquark6046:
Fix #867
Fix https://github.com/AdguardTeam/AdguardFilters/issues/168277

### DIFF
--- a/filters/ThirdParty/filter_122_FanboysAnnoyances/metadata.json
+++ b/filters/ThirdParty/filter_122_FanboysAnnoyances/metadata.json
@@ -12,5 +12,5 @@
     "purpose:annoyances",
     "reference:101"
   ],
-  "trustLevel": "low"
+  "trustLevel": "high"
 }


Alex-302 (approved):

AdamWr (approved):

krystian3w:
See too: #866

PR #871, opened 2023-12-13
Disable Web Annoyances Ultralist

piquark6046:
I think that Web Annoyances Ultralist can be considered as an obsoleted filters list because it causes many incorrect blocking without maintenance and no related commits.
 - https://github.com/yourduskquibbles/webannoyances/issues/516

### DIFF
--- a/filters/ThirdParty/filter_201_WebAnnoyancesUltralist/metadata.json
+++ b/filters/ThirdParty/filter_201_WebAnnoyancesUltralist/metadata.json
@@ -11,5 +11,6 @@
   "tags": [
     "purpose:annoyances"
   ],
-  "trustLevel": "high"
+  "trustLevel": "high",
+  "disabled": true
 }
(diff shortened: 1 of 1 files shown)


AdamWr (approved):

filters/ThirdParty/filter_201_WebAnnoyancesUltralist/metadata.json
```diff
@@ -11,5 +11,6 @@
   "tags": [
     "purpose:annoyances"
   ],
-  "trustLevel": "high"
+  "trustLevel": "high",
+  "disabled": true
```
Alex-302, line 15: Please also change the name like here
https://github.com/AdguardTeam/FiltersRegistry/blob/7638091b16ef2754cabdee61f09735abc5b0ccd2/filters/filter_12_Safari/metadata.json#L3

Alex-302 (changes requested):

Alex-302 (approved):

PR #893, opened 2024-01-10
Improve removeRulePatterns

AdamWr:
Do not remove `$removeparam` modifier if its value contains part of other modifiers.
Related to - https://github.com/AdguardTeam/AdguardFilters/issues/170206#issuecomment-1883912624


zloyden (approved):

slavaleleka (approved):

AdamWr:
As far as I can see, `||deezer.com^$removeparam=deferredFl` is in filter list for uBo - https://filters.adtidy.org/extension/ublock/filters/17.txt, so if it doesn't work, then there is probably another reason for that.

PR #911, opened 2024-03-09
Filters that break web pages.

rnixs:
Blocking those domains will break the webpage.
We need a different response.

Please refer to the issue
- https://github.com/AdguardTeam/AdguardBrowserExtension/issues/2737


piquark6046:
Fixed by https://github.com/AdguardTeam/AdGuardSDNSFilter/pull/1649.

PR #918, opened 2024-03-24
fix https://github.com/AdguardTeam/AdguardFilters/issues/175556

BlazDT:
https://github.com/AdguardTeam/AdguardFilters/issues/175556

Their filter dev is known not to unblock urls.

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -442,5 +442,11 @@ $~object-subrequest
 ! End: EasyList Czech and Slovak
 !###########################
 !
+!### Peter Lowe's Blocklist ###
+! https://github.com/AdguardTeam/AdguardFilters/issues/175556
+||tags.tiqcdn.com^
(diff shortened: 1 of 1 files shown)


filters/exclusions.txt
```diff
@@ -442,5 +442,11 @@ $~object-subrequest
 ! End: EasyList Czech and Slovak
 !###########################
 !
+!### Peter Lowe's Blocklist ###
+! https://github.com/AdguardTeam/AdguardFilters/issues/175556
+||tags.tiqcdn.com^
```
zloyden, line 447: > Their filter dev is known not to unblock urls.

It's by design. I don't think that it should be fixed on our side. We are only responsible for AdGuard filters.

BlazDT: Then all third-party exclusions would have to be removed.

Alex-302: In other filters breakage is not acceptable.

PR #930, opened 2024-04-03
update compiler to v1.1.129

slavaleleka:
add `$header` modifier support

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "node": ">=18"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


AdamWr:
It seems that build is failing on `EasyList`:
```
Unable to optimize filter 101 with configuration[~=50%, min=30%, max=60%],calculated = 62.00%! Filter rules count: 72665. Optimized rules count: 45177.
```
but I guess it's not related to changes in this PR.

AdamWr (approved):

zloyden (approved):

Issue #929, opened 2024-04-03
Do not build Online Malicious URL Blocklist for AdGuard for iOS


slavaleleka:
### Prerequisites

- [X] This problem is not related to filtering issues. Otherwise, you should read this [article](https://kb.adguard.com/technical-support/reporting-tool);
- [ ] You're using an up-to-date version of the problematic filter;

### Problem description

The filter is not used by AdGuard for iOS, it should not be built for that platform at all

### Proposed solution

use `platformsExcluded` to avoid filter build for specific platform

### Privacy

- [X] I agree to follow this condition

zloyden:
https://github.com/AdguardTeam/FiltersRegistry/pull/935#issuecomment-2039254276.

PR #935, opened 2024-04-04
Disable Online Malicious URL Blocklist for iOS

Alex-302:
https://github.com/AdguardTeam/FiltersRegistry/issues/929


zloyden (approved):

ameshkov:
We should've done it back then, now it does not make much sense unfortunately

PR #959, opened 2024-06-20
Support `ext_chromium_mv3` platform

slavaleleka:
I've added
- `EXTENSION_CHROMIUM_MV3` to `custom_platforms.js` 
- `ext_chromium_mv3` to `platformsExcluded` for multiple filters so for chromium mv3 there will be built:
  - all recommended filters 
  - all security filters (because there is no recommended one in the security group)
  - AdGuard Experimental filter 
  - AdGuard URL Tracking filter

### DIFF
--- a/filters/ThirdParty/filter_101_EasyList/metadata.json
+++ b/filters/ThirdParty/filter_101_EasyList/metadata.json
@@ -11,5 +11,8 @@
   "tags": [
     "purpose:ads"
   ],
+  "platformsExcluded": [
+    "ext_chromium_mv3"
+  ],
   "trustLevel": "low"
 }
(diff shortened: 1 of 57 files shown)


ameshkov:
@slavaleleka can we do this in FiltersCompiler's default config first?

slavaleleka:
> can we do this in FiltersCompiler's default config first?

I will update the compiler's config after a successful review of this pr. anyway a custom config has higher priority (and it can already be used for filters build now) so it is fine to update the compiler's config

maximtop (approved):

AdamWr (approved):

zloyden (approved):

piquark6046 (approved):

PR #967, opened 2024-07-02
update compiler to v1.1.141

slavaleleka:
https://github.com/AdguardTeam/FiltersCompiler/blob/v1.1.141/CHANGELOG.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "node": ">=18"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


AdamWr:
Is it okay that `Diff-Path` is removed?
<details><summary>Screenshot</summary>

image

</details>

slavaleleka:
> Is it okay that `Diff-Path` is removed?

when only `yarn build --include=2` is run, it is fine that `Diff-Path` filter list header is removed.
it returns after `yarn build:patches --time=3600 --resolution=s --include=2` is executed

AdamWr (approved):

PR #989, opened 2024-08-14
[WIP] update compiler to v1.1.144

slavaleleka:
https://github.com/AdguardTeam/FiltersCompiler/blob/master/CHANGELOG.md


filters/ThirdParty/filter_214_ABPVNList/filter.txt
```diff
@@ -4,7 +4,6 @@
 ! id: D4028CDD-3D39-4624-ACC7-8140F4EC3888
 ! Title: ABPVN List
 ! Last modified: 12 Aug 2024 14:55:10 UTC+7
-! Diff-Path: patches/20240812145510817.patch#abpvn_content_blocker
```
slavaleleka, line 7: I had to remove it manually once. this header tag should not appear in the future

Alex-302 (approved):

filters/ThirdParty/filter_214_ABPVNList/filter.txt
105th: @slavaleleka 
maybe it should be done by compiler to ensure that we delete Diff-Path from all thrird-party filters and to ensure that new checksums will be recalculated?

slavaleleka: it's been one case yet, and I'd rather keep as is for now

slavaleleka: new checksums shall be recalculated anyway

PR #994, opened 2024-08-28
Considering excluding the MV3 extension from 2 of my lists

DandelionSprout:
So, well, it's not exactly a secret that I'm hostile to the overall concept of MV3, and have been for around 3 years by now.

So I've gradually and loosely thought about excluding 2 more of my lists from AG's dedicated MV3 extension, presuming people would chip in with their views on it. If the consensus would end up as "It'd be ridiculous to exclude regional lists from the MV3 extension" or similar wordings, then I'd accept such a verdict.

### DIFF
--- a/filters/ThirdParty/filter_249_NorwegianList/metadata.json
+++ b/filters/ThirdParty/filter_249_NorwegianList/metadata.json
@@ -17,5 +17,8 @@
     "lang:is",
     "lang:fo"
   ],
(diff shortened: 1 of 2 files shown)


DandelionSprout:
Sounds good enough to me. 🌞

Alex-302:
So leave it as is - don't exclude the filter?

PR #999, opened 2024-09-19
Fix mobile filter

alxxjd:
There is a problem with dogdrip.net , wfmz, etc on the mobile Adguard. After erasing them one by one and testing them all, all the filters that have been removed have an effect.

```
solve below issues
- https://github.com/AdguardTeam/AdGuardForSafari/issues/1020
- https://github.com/AdguardTeam/AdguardFilters/issues/189115
```


mason945:
The issue is resolved by exempting the removed filters here.

Yuki2718:
It's a wrong place to open PR, but I don't think literally ALL - I restored those having no problem.

PR #1019, opened 2024-10-01
Fix porngo.xxx

jellizaveta:
Seems just images and content
https://github.com/AdguardTeam/AdguardFilters/issues/190067


Alex-302:
Isn't the site - is ad? Strange link.

jellizaveta:
@Alex-302 looks like an aggregator of content from other sites

PR #1020, opened 2024-10-02
Exclude $redirect-rule from filters for MV3 extension

jellizaveta:
https://github.com/AdguardTeam/FiltersCompiler/issues/231

### DIFF
--- a/scripts/build/custom_platforms.js
+++ b/scripts/build/custom_platforms.js
@@ -604,7 +604,11 @@ module.exports = {
         'path': 'extension/chromium-mv3',
         'expires': '10 days',
         'configuration': {
-            'removeRulePatterns': CHROMIUM_BASED_EXTENSION_PATTERNS,
+            'removeRulePatterns': [
+                ...CHROMIUM_BASED_EXTENSION_PATTERNS,
+                // https://github.com/AdguardTeam/FiltersCompiler/issues/231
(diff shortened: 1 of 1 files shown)


scripts/build/custom_platforms.js
```diff
@@ -604,7 +604,11 @@ module.exports = {
         'path': 'extension/chromium-mv3',
         'expires': '10 days',
         'configuration': {
-            'removeRulePatterns': CHROMIUM_BASED_EXTENSION_PATTERNS,
+            'removeRulePatterns': [
+                ...CHROMIUM_BASED_EXTENSION_PATTERNS,
+                // https://github.com/AdguardTeam/FiltersCompiler/issues/231
```
Alex-302, line 609: Maybe comments are redundant? It is (will be) described in the syntax docs.

zloyden (approved):

scripts/build/custom_platforms.js
slavaleleka: the main reason why this file was made `.js` (instead of `.json`) is to have an ability to add comments. so I believe such comments are fine

Alex-302 (approved):

scripts/build/custom_platforms.js
Alex-302: Every line must be commented even if all info exists in docs?)

PR #1056, opened 2025-02-13
AG-29276 mark filter 14 and 15 as deprecated

slavaleleka:
– 14 — AdGuard Annoyances filter
– 15 — AdGuard DNS filter, related to #800

https://github.com/AdguardTeam/FiltersRegistry#metadata

### DIFF
--- a/filters/filter_15_DnsFilter/metadata.json
+++ b/filters/filter_15_DnsFilter/metadata.json
@@ -4,6 +4,7 @@
   "description": "Filter composed of several other filters (AdGuard Base filter, Social Media filter, Tracking Protection filter, Mobile Ads filter, EasyList and EasyPrivacy) and simplified specifically to be better compatible with DNS-level ad blocking.",
   "timeAdded": 1404115015843,
   "homepage": "https://adguard.com/kb/general/ad-filtering/adguard-filters/",
+  "deprecated": true,
   "expires": "5 days",
   "displayNumber": 3,
   "groupId": 6,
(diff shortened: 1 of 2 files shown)


zloyden (approved):

piquark6046 (approved):

filters/filter_15_DnsFilter/metadata.json
```diff
@@ -4,6 +4,7 @@
   "description": "Filter composed of several other filters (AdGuard Base filter, Social Media filter, Tracking Protection filter, Mobile Ads filter, EasyList and EasyPrivacy) and simplified specifically to be better compatible with DNS-level ad blocking.",
   "timeAdded": 1404115015843,
   "homepage": "https://adguard.com/kb/general/ad-filtering/adguard-filters/",
+  "deprecated": true,
```
AdamWr, line 7: AdGuard DNS filter is still available in AdGuard for Android and MV2 extensions, why it should be marked as deprecated?

slavaleleka: HostlistRegistry's version of AdGuard DNS filter should be used for AdGuard for Android, and it may no longer be needed for the browser extension

AdamWr: AdGuard DNS filter is still available in AdGuard for Android in two places, one in "general" filters, let's say for usual ad blocking, and second in DNS protection. Are both from "HostlistRegistry"?

slavaleleka: > deprecated — optional, boolean, filter is deprecated and should not be used by the products, but filter is still available, i.e. not removed.

> Are both from "HostlistRegistry"?

I don’t know which version of the DNS filter is used in AdGuard for Android. However, this should not be a problem since the filter will remain available. It is marked as deprecated so that the product can decide whether to use it or not. Additionally, the "general" filters are related to content filtering, not DNS filtering—which is another reason why deprecating it makes sense

AdamWr (approved):

Alex-302 (approved):

PR #1064, opened 2025-03-13
update diff-builder to 1.1.0 to use system diff utility for patches

105th:
We released a new `diff-builder`, version `1.1.0`, which now uses the system `diff` tool to help prevent problems with creating large patches.

And skip some redundant logs.

### DIFF
--- a/scripts/build/patches.js
+++ b/scripts/build/patches.js
@@ -79,8 +79,13 @@ const main = async () => {
 
             const filename = path.basename(file);
 
+            // Just for optimization, skip files that are not in the format of
+            // "{filterId}[_optimized|_without_easylist].txt"
             if (!/\d+(_optimized|_without_easylist)?\.txt/.test(filename)) {
-                console.log(`Skipped generating patch for: ${file}`);
+                // Skip printing logs for non ".txt" files as redundant.
+                if (fileHasTxtExtension) {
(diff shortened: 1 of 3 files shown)


scripts/build/patches.js
```diff
@@ -79,8 +79,13 @@ const main = async () => {
 
             const filename = path.basename(file);
 
+            // Just for optimization, skip files that are not in the format of
+            // "{filterId}[_optimized|_without_easylist].txt"
             if (!/\d+(_optimized|_without_easylist)?\.txt/.test(filename)) {
-                console.log(`Skipped generating patch for: ${file}`);
+                // Skip printing logs for non ".txt" files as redundant.
```
Alex-302, line 85: > // Skip printing logs for non ".txt" files as redundant.

Some filters don't have an extension
`@include "https://www.i-dont-care-about-cookies.eu/abp/"`
`@include "https://raw.githubusercontent.com/DandelionSprout/adfilt/master/AnnoyancesList"`

105th: @Alex-302 This script checks against filenames only in the file system, nothing more, for proper selection of files to which patches should be generated.

maximtop (approved):

AngSolids (approved):

Issue #1091, opened 2025-06-16
Fix failing builds of 3p filters


105th:
### Prerequisites

- [x] This problem is not related to filtering issues. Otherwise, you should read this [article](https://kb.adguard.com/technical-support/reporting-tool);
- [x] You're using an up-to-date version of the problematic filter;

### Problem description

During last successful build of 3p filters (https://github.com/AdguardTeam/FiltersRegistry/actions/runs/15648473950) header `Diff-Path` for filter 250 was deleted.
It happened because `diff-builder` library was validating patch and after applying it - it was not equal to new filter - https://github.com/AdguardTeam/FiltersRegistry/actions/runs/15648473950/job/44089768836#step:6:37219. So library just skipped adding `Diff-Path` to the new filter and we left it without `Diff-Path`.

Now this causing an error for fresh versions, since current versions of filter 250 already contain returned `Diff-Path` header and library creates new empty patch for them.

After that, post-validation will throw an error since we are not allow two empty patches in a directory for a proper work of update by patches in a chain.

### Proposed solution

Delete empty patches for 250 and re-run build for 3p filters.

### Privacy

- [x] I agree to follow this condition

105th:
Fixed via settings correct encoding for diff generation

PR #1092, opened 2025-06-16
#1091: delete patches which will never be filled for correct future builds

105th:
For correct generation future patches we need to delete patches of 250 filters where we occasionally deleted `Diff-Path`.


105th:
All platforms will wait until Filter expiration date (currently - 10 days, started from 14 of June) for next full redownload filter.

PR #1096, opened 2025-06-20
Fix intel.com

AdamWr:
Fix https://github.com/AdguardTeam/AdguardFilters/issues/207905 

Fixes problem with downloading files on `intel.com`.

File cannot be downloaded when `dogo.intel.com` is blocked (it comes from https://github.com/AdguardTeam/cname-trackers/blob/master/data/trackers/adobe-experience-cloud-(formerly-omniture).txt).

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -50,6 +50,8 @@ $popup,~third-party
 !###################
 !
 !### cname-trackers problems ###
+! https://github.com/AdguardTeam/AdguardFilters/issues/207905
+||dogo.intel.com^
 ! https://github.com/AdguardTeam/AdguardFilters/issues/194839
 ||data-3cd8fb3825.kicker.de^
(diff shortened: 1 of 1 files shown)


BlazDT (approved):

Alex-302:
Should not it be unblocked in DNS filter?
Like here https://github.com/AdguardTeam/AdGuardSDNSFilter/pull/1912/files

Alex-302:
CC: @hagezi I see same rule in your filters. Are users not complaining?

hagezi:
@Alex-302 So far there have been no issues, I'll have a look.

Alex-302:
> But perhaps it should be unblocked in two places?

Hm. Let's check if it disappears in the DNS filter after next compile of FiltersRegistry.

Alex-302 (approved):

zloyden (approved):

PR #1097, opened 2025-06-21
Fix fanboynz's automatic script issue

piquark6046:
To avoid any possible breakage in the future.
 - https://github.com/easylist/easylist/commit/14c14e50ef703cbfef435cb48351a461caa2367f
 - https://github.com/AdguardTeam/AdguardFilters/issues/207951
 - https://github.com/AdguardTeam/AdguardFilters/issues/207950
 - https://github.com/easylist/easylist/commit/0924d8a5781d5592efc570192cf7cc8f4f4fa1a4#diff-1298f15fc987f32c8ee53ef08b4582d0861e6fecced9888537d7929b781c9ed2R6968

### DIFF
--- a/filters/exclusions.txt
+++ b/filters/exclusions.txt
@@ -85,6 +85,8 @@ $popup,~third-party
 !
 !### Easylist ###
 !
+! Fanboynz's automatic script issue
+||discord.com^
 ! We have ##.ad-unit:not(.textads) in our filters
 /^(~.*)?##\.ad-unit$/
 ! Causes many problems, especially in Safari


BlazDT (approved):

piquark6046:
Then, we can add another case that blocks any normal service's domain into the section in the future

Yuki2718:
I mean `discord.com` was already excluded in his script, what is the point of adding exclusion every time AFTER it became an issue?

piquark6046:
> I mean `discord.com` was already excluded in his script, what is the point of adding exclusion every time AFTER it became an issue and fixed in the script?

I believe that the 2nd accident should not be happended because adding `||discord.com^` can be removed at the postprocessing (like checking well-known web service's domain) of fanboynz's automatic script.

Yuki2718:
> Discord.com got blocked several times in EasyList. I guess therefore to ensure it is not blocked anymore in the future.

Seems you're right about this: https://github.com/easylist/easylist/commit/0924d8a5781d5592efc570192cf7cc8f4f4fa1a4 So maybe excluding discord still makes sense.

PR #1106, opened 2025-07-07
Fix AdGuard Base filter optimization config

Alex-302:


slavaleleka (approved):
PR title should be changed

PR #1109, opened 2025-07-17
Fix map.naver.com

piquark6046:
https://github.com/AdguardTeam/AdguardFilters/issues/204843#issuecomment-3077532042

```typescript
window.MutationObserver = new Proxy(window.MutationObserver, {
    construct(Target: typeof MutationObserver, Args: ConstructorParameters<typeof MutationObserver>) {
        let Stringified = Args[0].toString()
        if (Stringified.includes('cs:') && Stringified.includes('gl:') && Stringified.includes('ne:') && Stringified.includes('dg:')) {
            return new Promise(() => {})
        }
        return Reflect.construct(Target, Args)
    }
})
```

https://github.com/AdguardTeam/Scriptlets/issues/461#issuecomment-3081859146

PR #1110, opened 2025-07-17
Revert "Fix map.naver.com"

piquark6046:
Reverts AdguardTeam/FiltersRegistry#1109
https://github.com/AdguardTeam/FiltersRegistry/pull/1109#issuecomment-3082853303


piquark6046:
By the way, I think that It is an issue about region of the IP address.
Maybe can connecting to a South Korean IP address help?

PR #1118, opened 2025-07-31
Update compiler

jellizaveta:


.eslintignore
```diff
@@ -1 +1,3 @@
 platforms/
+scripts/build/
```
scripthunter7: why ignored?

package.json
```diff
@@ -4,52 +4,54 @@
...
+        "build": "node -r @swc-node/register scripts/build/build.js",
         "build:patches": "node scripts/build/patches.js",
         "validate": "yarn validate:platforms && yarn validate:locales",
-        "validate:platforms": "node scripts/validation/validate_platforms.js",
-        "validate:locales": "node scripts/validation/validate_locales.js",
-        "update-wildcard-domains": "node -r @swc-node/register scripts/wildcard-domain-processor update-wildcard-domains",
-        "expand-wildcard-domains": "node -r @swc-node/register scripts/wildcard-domain-processor expand-wildcard-domains",
+        "validate:platforms": "node -r @swc-node/register scripts/validation/validate_platforms.js",
+        "validate:locales": "node -r @swc-node/register scripts/validation/validate_locales.js",
+        "update-wildcard-domains": "tsx scripts/wildcard-domain-processor update-wildcard-domains",
```
scripthunter7, line 20: maybe it would be better to use only tsx or swc-node/register

package.json
```diff
@@ -4,52 +4,54 @@
...
+        "expand-wildcard-domains": "tsx scripts/wildcard-domain-processor expand-wildcard-domains",
         "compress": "node scripts/repository/compress.js"
     },
     "engines": {
-        "node": ">=18"
+        "node": ">=22"
```
scripthunter7, line 25: which part of the code requires at least node 22?

scripts/validation/validate_platforms.js
```diff
@@ -1,9 +1,13 @@
...
+import fs from 'fs';
+import { fileURLToPath } from 'url';
+import { validateJSONSchema } from '@adguard/filters-compiler';
+import { PATCH_EXTENSION } from '@adguard/diff-builder';
 
-const { FOLDER_WITH_NEW_FILTERS } = require('../build/constants');
+import { FOLDER_WITH_NEW_FILTERS } from '../build/constants.js';
+
+const filename = fileURLToPath(import.meta.url);
+const dirname = path.dirname(filename);
```
scripthunter7: prefer __filename and __dirname, please also check in other places

scripts/wildcard-domain-processor/domain-extractor.ts
```diff
@@ -26,7 +30,7 @@ export function extractModifierDomains(modifier: Modifier): {
         return [];
     }
 
-    const domains = DomainListParser.parse(modifier.value.value, agtree.PIPE_MODIFIER_SEPARATOR);
+    const domains = DomainListParser.parse(modifier.value.value, {}, modifier.start, PIPE_MODIFIER_SEPARATOR);
```
scripthunter7: you can import default parser options from agtree instead of empty object

scripts/wildcard-domain-processor/domain-extractor.ts
```diff
@@ -91,7 +95,10 @@ function extractNetworkRuleDomains(ast: agtree.NetworkRule): string[] {
 export function extractRuleDomains(ast: AnyRule): string[] {
     switch (ast.category) {
         case RuleCategory.Network:
-            return extractNetworkRuleDomains(ast);
+            if ('pattern' in ast && 'exception' in ast) {
```
scripthunter7: maybe it would be better to check rule type to exclude host rules here

scripts/wildcard-domain-processor/index.ts
```diff
@@ -34,7 +35,7 @@ program
     });
 
 // Only run the command-line interface if the script is executed directly
-if (require.main === module) {
+if (process.argv[1] === fileURLToPath(import.meta.url)) {
```
scripthunter7: please prefer __dirname const instead

scripts/wildcard-domain-processor/wildcard-expander.ts
```diff
@@ -1,24 +1,29 @@
...
-    RuleParser,
 } from '@adguard/agtree';
 
-import { findFilterFiles, readFile, writeFile } from './file-utils';
-import { type AliveWildcardDomains } from './wildcard-domains-updater';
-import { DOMAIN_MODIFIERS } from './domain-extractor';
-import { utils } from './utils';
-import { updateContentChecksum } from '../checksum';
+import { PIPE_MODIFIER_SEPARATOR, NEGATION_MARKER } from '@adguard/agtree/utils';
+
```
scripthunter7: remove redundant empty lines

scripthunter7 (changes requested):
generally looks good to me, nice work. just left some comments, please check them

package.json
jellizaveta: `@adguard/agtree@3.2.2 expected version ">=22"`

scripthunter7: i think its too strict, probably we can reduce it in agtree as well

scripts/build/custom_platforms.js
```diff
@@ -821,3 +821,5 @@ module.exports = {
         },
     },
 };
+
+export default PLATFORMS_CONFIG;
```
slvvko: can we avoid _default_ export and use only _named_ one?

scripts/wildcard-domain-processor/index.ts
```diff
@@ -1,9 +1,9 @@
 /* eslint-disable no-console */
 import { Command } from 'commander';
 import path from 'path';
-
-import { updateWildcardDomains } from './wildcard-domains-updater';
-import { expandWildcardDomains } from './wildcard-expander';
+import { fileURLToPath } from 'url';
+import { updateWildcardDomains } from './wildcard-domains-updater.js';
```
slvvko, line 5: I suppose `.js` endings in imports should be avoided in the `.ts` files

slvvko (approved):

package.json
slvvko: I prefer not to decrease it in agtree,
especially when node.js v18 is no longer maintained now:
https://nodejs.org/en/about/previous-releases

let's keep it 22

zloyden (approved):

AdamWr (approved):

scripthunter7 (approved):

PR #1123, opened 2025-08-14
Fix idlixku.com

piquark6046:
Fix https://github.com/AdguardTeam/AdguardFilters/issues/211307


Alex-302:
Can be handled in Easylist.

PR #1126, opened 2025-08-24
Reintroduce Web Annoyances Ultralist

lsomm:
Hi, after the [original Web Annoyances Ultralist](https://github.com/yourduskquibbles/webannoyances) became unmaintained, it got removed in #871.

Two years and 459 commits later, we have a [well-established fork](https://github.com/LanikSJ/webannoyances) with many new commits, fixes, etc, a small but established community and a quick maintainer.

It would be nice to add us back again, so that more people can benefit from the (now bugfixed and updated) list, as well as to increase attention to the fork, so that more people feel like contributing, who might have in the past. 

This PR is adding the updated fork again to the list of maintained third party sources. I have just tried to update the URLs and revert the changes done in #871. If I did anything wrong, I apologise.

I thought just asking with a quick PR is the easiest way; I hope this is not too blunt.

### DIFF
--- a/filters/ThirdParty/filter_201_WebAnnoyancesUltralist/metadata.json
+++ b/filters/ThirdParty/filter_201_WebAnnoyancesUltralist/metadata.json
@@ -1,20 +1,18 @@
 {
   "filterId": 201,
-  "name": "(Obsolete) Web Annoyances Ultralist",
+  "name": "Web Annoyances Ultralist",
   "description": "Blocks annoying web elements and reclaims lost screen real estate.",
   "timeAdded": 1560248829763,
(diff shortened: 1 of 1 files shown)


piquark6046:
It seems that the filters list does not have enough GitHub stars to add any filters list.

Alex-302:
We'll keep an eye on it.

ShinyaReverse:
Regarding the page display issue, the layout can be disrupted, causing overlapping text and elements, as shown in the attached screenshot. 
Screenshot_20260131-023809 Chrome


Other issues, such as the collapsible menu not expanding, can also occur. 
It's difficult to generalize this issue, so I think the only solution is to fix it individually as soon as it's discovered.

Issue #954, opened 2024-05-22
Add «Dandelion Sprout's Anti-Malware List»


DandelionSprout:
### Prerequisites

- [X] I checked the documentation and understood it;
- [X] I checked to make sure that this issue has not already been filed;

### Problem description

Hoping to make my anti-malware list recognised as a major security list once and for all, alongside the small handful of other lists in that field that doesn't really cover much (in particular Online Malicious URL Blacklist).

### Proposed solution

Add https://raw.githubusercontent.com/DandelionSprout/adfilt/master/Alternate%20versions%20Anti-Malware%20List/AntiMalwareAdGuard.txt to https://github.com/AdguardTeam/FiltersRegistry/tree/master/filters/ThirdParty

### Additional information

I attempted to figure out how to create a full pull request for this, but unfortunately the repo system for lists was incomprehensible to me. 😓

The list is also covered by Dandelion Sprout's Annoyances List, so a note similar to that used for "EasyList" towards "AdGuard Base" (that the former is already covered by the latter), should be added for Anti-Malware List as well as for Legitimate URL Shortener towards Dandelion Sprout's Annoyances List.

DandelionSprout:
### Additional notes added at a later time than the OP

The list is to the best of my current knowledge the only dedicated anti-malware list to use AdGuard-specific syntax, in particular `$network`. However, it dawned on me only just now that I'll need to do dual-coverage of the paid AdGuard versions and the Browser Extension versions (by using both `250.250.250.250$network` *and* `||250.250.250.250^$all`). I can handle that later this week.

The way the list is made has some similarities to uBlock Filters Badware Risks, in that both lists are based on live experiences from people who browse the internet normally (whereas alleged security firms out there just base them on very poorly performed filescans). The main difference (that I'm currently aware of) between my list and Badware Risks is that I usually cover all domains found on bad IPs (occasionally reaching 3- or 4-digit amounts of domains), while they cover the most common malware domains that users encounter.

krystian3w:
A little progress:

https://github.com/AdguardTeam/FiltersRegistry/pull/1081

zloyden:
The filter has been added.

PR #1143, opened 2025-10-22
Added `Dandelion Sprout's Anti-Malware List`

zloyden:
https://github.com/AdguardTeam/FiltersRegistry/issues/954


filters/ThirdParty/filter_259_Dandelion_Sprout's_Anti-Malware_List/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+  "filterId": 259,
+  "name": "Dandelion Sprout's Anti-Malware List",
+  "description": "This list goes the extra kilometer to prevent more malware than other mainstream anti-malware lists. It blocks domains and domain patterns used in malware redirection trains and in domain parking schemes, blocks sponsored Windows PUP nags on PC guide articles, uses mass blocking of domains belonging to bad IPs, and has many other subcategories that give it a solid advantage over similar lists out there.",
```
Alex-302: @DandelionSprout Maybe it would be better to make the description more concise? I mean only metadata in FiltersRegistry, without changing the filter file itself.

How current description will look in the extension

Image

DandelionSprout: The best I could come up with tonight:

*"Uses AdGuard syntax to prevent more malware than most other major anti-malware lists. Blocks domains and domain patterns used in malware redirection trains, IP addresses that are solely used by those trains, PUP nags, and a few scammers. Already included in Dandelion Sprout's Annoyances List."*

Crowdin translators should be made aware that "trains" refers to chains of multiple rapid redirects; so that if train (as in the vehicle) analogies don't work in their languages, they are free to use other analogies; and also that it has nothing to do with any sort of training.

Alex-302: Maybe a bit shorter? It's your version, but it takes up one line less in the current extension design.

> Blocks more malware than most other major anti-malware lists - domains and URL patterns used in malware redirection chains, IP addresses that are solely used by malware, PUP nags, and a few scammers. Already included in Dandelion Sprout's Annoyances List.

DandelionSprout: Okay, sure thing.

PR #1152, opened 2025-11-20
Update exclusions.txt

krystian3w:
First is disabled by comment; additionally, the space has disappeared and new TLDs and CSS negation have appeared there:

https://github.com/AdguardTeam/FiltersRegistry/blob/f4c27530b267169443350295c3e0acba74346ad7/platforms/extension/android-content-blocker/filters/216.txt#L252

Nike rules in that form no longer work and do not exist on the lists.

The [gf](https://github.com/FiltersHeroes/PolishAnnoyanceFilters/blob/33260f49c1e8718359e42d116168d67c1ab6866b/scriptlets/scriptlets.js#L5-L18) scriptlet rule is no longer needed on meczyki (push without damage to thumbnails blocks `meczyki.pl##+js(set, Notification.requestPermission, falseFunc)` in latest Firefox/Chrome), in comment mode for a long time.


AdamWr (approved):

Alex-302 (approved):
Thanks.

Issue #1153, opened 2025-11-25
Add AWAvenue Ads Rule as an upstream source for AdGuard Chinese Filter


TG-Twilight:
### Prerequisites

- [x] I checked the documentation and understood it;
- [x] I checked to make sure that this issue has not already been filed;

### Problem description

AdGuard Chinese Filter currently focuses on improving the ad-blocking experience for Chinese-speaking users, but many region-specific ads, tracking parameters, mobile ads pop-ups and app-based ad formats are still better covered by community-maintained lists that update more frequently and respond more quickly to new advertising patterns.

### Proposed solution

AWAvenue Ads Rule is an actively maintained, fast-updating Chinese-focused ad filter list with a high user adoption rate. It includes rules optimized for major mobile apps, embedded ads. Many AdGuard users in the Chinese community already use this list manually, so adding it as an upstream source helps improve coverage and benefits the entire user base.

Integrating this filter can effectively enhance the accuracy and update responsiveness of AdGuard Chinese Filter.

PR #1154, opened 2025-11-26
Adding the `AWAvenue Ads Rule` to the `AdGuard Chinese` filter

zloyden:
https://github.com/AdguardTeam/FiltersRegistry/issues/1153

### DIFF
--- a/filters/filter_224_Chinese/template.txt
+++ b/filters/filter_224_Chinese/template.txt
@@ -38,3 +38,7 @@
 !--------------- EasyList China rules --------------------------------------!
 !---------------------------------------------------------------------------!
 @include "https://easylist-downloads.adblockplus.org/easylistchina.txt" /stripComments /exclude="../exclusions.txt"
+!---------------------------------------------------------------------------!
+!--------------- AWAvenue Ads Rule -----------------------------------------!
+!---------------------------------------------------------------------------!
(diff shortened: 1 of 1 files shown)


zloyden:
@TG-Twilight please also take a look at this,

AdamWr:
One thing, as far as I can see, it's multipurpose list and it's blocking `||fundingchoicesmessages.google.com^`.
Some websites are broken when it's blocked.
I guess we should add exception for domains from this rule:
```adblock
ot.gr,tvi.iol.pt,cnnturk.com,news.bg,lifestyle.bg,topsport.bg,hurriyet.com.tr,kanald.com.tr#%#//scriptlet('trusted-click-element', '.fc-consent-root .fc-footer-buttons-container > .fc-footer-buttons > .fc-cta-consent.fc-primary-button')
```
once this PR is merged.
At least for `hurriyet.com.tr` (video player does not work here - `https://www.hurriyet.com.tr/spor/videolar/diger/video-montelladan-kura-sonrasi-ilk-aciklama-43026245` when `fundingchoicesmessages.google.com` is blocked), I'm not sure about other domains.

zloyden:
@AdamWr that's a good point. I think it shouldn't be blocked at all, as it goes to the AdGuard Chinese filter.

PR #1157, opened 2025-12-11
Restored redirect/removeparam rules for mv3 platforms

Alex-302:
Don't merge until approved by @slvvko

### DIFF
--- a/scripts/build/custom_platforms.js
+++ b/scripts/build/custom_platforms.js
@@ -658,10 +658,7 @@ export const CUSTOM_PLATFORMS_CONFIG = {
         'path': 'extension/chromium-mv3',
         'expires': '10 days',
         'configuration': {
-            'removeRulePatterns': [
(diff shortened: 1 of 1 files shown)


scripts/build/custom_platforms.js
```diff
@@ -658,10 +658,7 @@ export const CUSTOM_PLATFORMS_CONFIG = {
         'path': 'extension/chromium-mv3',
         'expires': '10 days',
         'configuration': {
-            'removeRulePatterns': [
-                ...CHROMIUM_BASED_EXTENSION_PATTERNS,
-                ...REDIRECT_MODIFIER_PATTERNS,
```
AdamWr, line 663: Should not `REDIRECT_MODIFIER_PATTERNS` be replaced with `REDIRECT_RULE_MODIFIER_PATTERNS`?
If I'm not wrong, `redirect-rule` modifier is not supported by MV3 extension.
The same below.

PR #1181, opened 2026-03-28
Fix #1180 Add the ability to build filters completely locally

Alex-302:
https://github.com/AdguardTeam/FiltersRegistry/issues/1180


scripts/build/strip-generated-meta.ts
maximtop, line 1: Let's start using TypeScript here as well. We're gradually adopting it across all our projects, and I can already see TypeScript files in this repo.

maximtop, line 1: Could we use `kebab-case` for the new filename here?
We are gradually standardizing on this naming style across projects, so for new files it would be better to move in that direction instead of extending the underscore-based pattern.

scripts/build/strip_generated_meta.js
```diff
@@ -0,0 +1,130 @@
...
+ * regardless of filter content changes, making file comparison noisy.
+ */
+const GENERATED_META_PREFIXES = [
+    '! Checksum:',
+    '! Diff-Path:',
+    '! TimeUpdated:',
+    '! Version:',
+];
+
+const isGeneratedMetaLine = (line) => GENERATED_META_PREFIXES.some((prefix) => line.startsWith(prefix));
```
maximtop: Could we avoid the one-liner here and keep it in the same style as the rest of the file? We usually document helper functions with JSDoc, so it would be better to turn this into a small named function with a short doc comment.

scripts/build/strip_generated_meta.js
```diff
@@ -0,0 +1,130 @@
+import fs from 'fs';
```
maximtop: `import fs from 'fs/promises'`

maximtop: It looks like this file only uses `fs.promises`. Could we import that directly instead of the whole `fs` module?

scripts/build/strip-generated-meta.ts
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * Strip generated metadata lines from a single filter file in-place.
+ *
+ * @param {string} filePath - Absolute path to the .txt filter file.
+ * @returns {Promise<boolean>} True if the file was modified, false otherwise.
+ */
+const stripGeneratedMeta = async (filePath) => {
+    const content = await fs.promises.readFile(filePath, 'utf8');
+    const lines = content.split('\n');
+    const filtered = lines.filter((line) => !isGeneratedMetaLine(line));
```
maximtop, line 39: These meta lines are only expected near the beginning of the file, so scanning and rebuilding the whole file looks unnecessary. We could limit the scan to the initial part of the file, for example the first 100 lines, instead of filtering every line.

scripts/build/strip_generated_meta.js
```diff
@@ -0,0 +1,130 @@
...
+ */
+const findTxtFiles = async (dir) => {
+    const entries = await fs.promises.readdir(dir, { withFileTypes: true });
+
+    const results = await Promise.all(entries.map(async (entry) => {
+        const fullPath = path.join(dir, entry.name);
+        if (entry.isDirectory()) {
+            return findTxtFiles(fullPath);
+        }
+        if (entry.isFile() && entry.name.endsWith('.txt')) {
```
maximtop: it would be good to add this .txt extension to the constant

scripts/build/strip_generated_meta.js
```diff
@@ -0,0 +1,130 @@
...
+    return true;
+};
+
+/**
+ * Recursively find all .txt files inside a given directory.
+ *
+ * @param {string} dir - Root directory to search.
+ * @returns {Promise<string[]>} Array of absolute paths to .txt files.
+ */
+const findTxtFiles = async (dir) => {
```
maximtop: This function looks like duplicated filesystem traversal logic. We already have a recursive file-finding helper in `scripts/utils/find_files.js`, so could we reuse that here instead of adding another implementation?

scripts/build/strip_generated_meta.js
```diff
@@ -0,0 +1,130 @@
...
+ */
+const findFiltersDirs = async (dir) => {
+    const entries = await fs.promises.readdir(dir, { withFileTypes: true });
+
+    const results = await Promise.all(entries.map(async (entry) => {
+        if (!entry.isDirectory()) {
+            return [];
+        }
+        const fullPath = path.join(dir, entry.name);
+        if (entry.name === 'filters') {
```
maximtop: add `'filters'` to the constant

scripts/build/strip-generated-meta.js
```diff
@@ -0,0 +1,130 @@
...
+
+    return results.flat();
+};
+
+/**
+ * Strip generated metadata lines from all .txt files inside all `filters/`
+ * directories found recursively under the given root.
+ *
+ * @param {string} rootDir - Root directory to search (e.g. `platforms/`).
+ * @returns {Promise<number>} Number of files actually modified.
```
maximtop: I do not think there is a real race condition here, but mutating `totalModified` inside parallel `Promise.all` callbacks is still a bit awkward. It would be cleaner to return per-directory counts and sum them after `Promise.all`, instead of relying on shared mutable state.

scripts/build/build.js
```diff
@@ -40,26 +46,128 @@ args.forEach((val) => {
     if (val.startsWith('--report=')) {
         rawReportPath = val.slice(val.indexOf('=') + 1).trim();
     }
+
+    if (val === '--use-cache') {
+        useCache = true;
+    }
+
+    if (val === '--generate-cache') {
```
maximtop: `yarn generate-cache` is exposed as a separate command, but internally it is still just another execution mode of `build.js`. From an architecture point of view, I am not sure cache generation belongs in the same script as the normal platform build flow. A dedicated script would make that separation clearer.

scripts/build/build.js
```diff
@@ -40,26 +46,128 @@ args.forEach((val) => {
...
 
+const SHADOW_TEMPLATE_CONTENT = '@include "./filter.txt"\n';
+
+/**
+ * Recursively find all `template.txt` files under the given directory.
+ *
+ * @param {string} dir - Root directory to search.
+ * @returns {Promise<string[]>} Array of absolute paths to `template.txt` files.
+ */
+const findTemplatePaths = async (dir) => {
```
maximtop: It looks like `findFiles` in `scripts/utils/find_files.js` could already cover at least part of this new traversal logic, for example the `template.txt` and `.txt` searches. Could we reuse it here instead of introducing another recursive file walker?

maximtop (changes requested):

scripts/build/strip-generated-meta.ts
Alex-302: Maybe later? It is completely different task.

Alex-302: and `custom_platforms.js`, `find_files.js`?

scripts/build/strip_generated_meta.js
Alex-302: Will fix.

scripts/build/strip-generated-meta.ts
Alex-302: The performance boost will be negligible, and the code will be more complex.
Also, if the filters are combined into a single large filter, the current version will process all occurrences.

scripts/build/strip_generated_meta.js
Alex-302: Will fix.

scripts/build/strip_generated_meta.js
Alex-302: Will fix.

scripts/build/build.js
Alex-302: reused find_files.js

scripts/build/build.js
Alex-302: This will lead to increased complexity due to code duplication. I think we can put this off until we refactor the code or completely rewrite it in TypeScript.

scripts/build/strip-generated-meta.ts
Alex-302: Ok, I got you now)

scripts/build/strip_generated_meta.js
Alex-302: Yep.

scripts/build/strip_generated_meta.js
Alex-302: Done.

scripts/build/strip-generated-meta.ts
Alex-302: It seems not now - used in the comment in FiltersCompiler
https://github.com/AdguardTeam/FiltersCompiler/blob/baa8636e6cac9c60b239520b35c68edea1883f51/src/main/platforms-config.js#L8

Alex-302: Done.

package.json
```diff
@@ -13,7 +13,10 @@
         "lint:md": "markdownlint **/*.md",
         "test": "vitest run",
         "build": "tsx scripts/build/build.js",
+        "build:local": "tsx scripts/build/build.js --use-cache",
         "build:patches": "node scripts/build/patches.js",
+        "strip-generated-meta": "tsx scripts/build/strip-generated-meta.ts",
```
maximtop, line 18: `yarn strip-generated-meta` points to `scripts/build/strip-generated-meta.ts`, but this file only exports `stripGeneratedMetaFromDir()` and never invokes it.

I verified locally that the command exits successfully without touching the generated files. Please add a CLI entrypoint here, or point the npm script to a small wrapper that actually calls the function.

scripts/build/build.js
```diff
@@ -70,27 +156,49 @@ const buildFilters = async () => {
...
     // the temp folder to create the first empty patches for future versions
-    if (initialRun) {
+    if (initialRun && !noPatchesPrepare) {
         // Make copy for future patches generation
         await fs.promises.cp(platformsPath, copyPlatformsPath, { recursive: true });
     }
+
+    // Strip generated metadata (Checksum, Diff-Path, TimeUpdated, Version)
+    // from compiled filter files so they don't pollute diff comparisons.
+    if (stripGeneratedMeta) {
```
maximtop, line 176: This strips metadata only from the new `platforms/` tree, but the old baseline may already have been copied to `temp/platforms/` above.

That means a later `yarn build:patches` can diff stripped output against an unstripped baseline and generate patch noise from volatile headers instead of real content changes. Please either make `--strip-generated-meta` imply `--no-patches-prepare`, reject the unsafe combination, or strip both trees consistently.

maximtop (changes requested):

scripts/build/strip-generated-meta.ts
```diff
@@ -0,0 +1,97 @@
...
+const hasTxtExtension = (p: string): boolean => p.endsWith(TXT_FILE_EXTENSION);
+
+/**
+ * Strip generated metadata lines from all .txt files inside all `filters/`
+ * directories found recursively under the given root.
+ *
+ * @param rootDir - Root directory to search (e.g. `platforms/`).
+ * @returns Number of files actually modified.
+ */
+export const stripGeneratedMetaFromDir = async (rootDir: string): Promise<number> => {
```
maximtop, line 98: Non-blocking: this first walks the tree to find every `filters/` directory and then does another recursive walk inside each one to collect `.txt` files. Since we already have a shared recursive file finder, could we avoid the extra traversal and gather the relevant files in one pass?

scripts/build/build.js
```diff
@@ -40,26 +47,105 @@ args.forEach((val) => {
...
+    }));
+
+    // eslint-disable-next-line no-console
+    console.log(`Prepared cached filters directory with ${templatePaths.length} shadow templates.`);
+};
+
 /**
  * Compiler entry point.
  */
 const buildFilters = async () => {
```
maximtop, line 112: This introduces several new execution paths and flag interactions (`--use-cache`, `--generate-cache`, `--no-patches-prepare`, `--strip-generated-meta`), but the PR does not add any focused coverage for them.

Could we extract the mode selection and argument handling into smaller helpers and add tests for the new combinations? The two blocker scenarios here are exactly the kind of regressions that are easy to miss without targeted coverage.

maximtop (changes requested):

scripts/build/build.js
slvvko, line 1: add tests for new flag combinations

slvvko (changes requested):

package.json
Alex-302: Fixed in https://github.com/AdguardTeam/FiltersRegistry/pull/1181/commits/7a10cdbfd63cd83e0932af3c51348412c2521896

scripts/build/build.js
Alex-302: Fixed in https://github.com/AdguardTeam/FiltersRegistry/pull/1181/changes/9c694a6ac0d25c6e996d0e6674eb007edcfe46ee

scripts/build/strip-generated-meta.ts
Alex-302: Fixed in https://github.com/AdguardTeam/FiltersRegistry/pull/1181/commits/a6919b5fd02019023cea9aaa341abed4148c2a69

scripts/build/build.js
Alex-302: tests were requested in https://github.com/AdguardTeam/FiltersRegistry/pull/1181#discussion_r3073357740

scripts/build/build.js
Alex-302: Added in https://github.com/AdguardTeam/FiltersRegistry/pull/1181/commits/aea8ac62625c9ec8154f5e84b404dd3c8e53fc4d

scripts/build/build.js
```diff
@@ -8,39 +8,53 @@ import {
...
+    generateCache,
+    noPatchesPrepare,
+    stripGeneratedMeta,
+} = flags;
+
+if (useCache && generateCache) {
+    // eslint-disable-next-line no-console
+    console.error('Error: --use-cache and --generate-cache are mutually exclusive.');
+    process.exit(1);
+}
```
maximtop: Non-blocking: this block is dead code now — `validateFlags(flags)` on line 32 already catches `useCache && generateCache` and exits at line 39. Safe to remove.

scripts/build/__tests__/build-config.test.ts
```diff
@@ -0,0 +1,170 @@
...
+        ]);
+
+        expect(flags.useCache).toBe(true);
+        expect(flags.noPatchesPrepare).toBe(true);
+        expect(flags.stripGeneratedMeta).toBe(true);
+    });
+
+    it('ignores unknown arguments', () => {
+        const flags = parseFlags(['--mode=unknown', '--foo']);
+        expect(flags.useCache).toBe(false);
```
maximtop, line 41: Non-blocking: `--generate-cache` is missing from this test. The other three boolean flags are covered but `generateCache` is not asserted.

```suggestion
    it('parses all boolean flags', () => {
        const flags = parseFlags([
            '--use-cache',
            '--generate-cache',
            '--no-patches-prepare',
            '--strip-generated-meta',
        ]);

        expect(flags.useCache).toBe(true);
        expect(flags.generateCache).toBe(true);
        expect(flags.noPatchesPrepare).toBe(true);
        expect(flags.stripGeneratedMeta).toBe(true);
    });
```

maximtop (approved):
All previous blockers addressed. CI green. Two non-blocking nits left (dead code in build.js, missing `--generate-cache` in boolean flags test). LGTM otherwise — clean extraction of build config, solid test coverage, good docs.

LGTM

slvvko (approved):

zloyden (approved):

scripts/build/build.js
Alex-302: Fixed https://github.com/AdguardTeam/FiltersRegistry/pull/1181/commits/02c6178f9615d111a1be5749667af10510c9547e

scripts/build/__tests__/build-config.test.ts
Alex-302: Fixed https://github.com/AdguardTeam/FiltersRegistry/pull/1181/commits/1c2438a048dc21608023795ed80ea05a037c4d7a

maximtop (approved):

PR #1203, opened 2026-06-16
Adding "Stevo's AI Blocklist"

zloyden:


zloyden:
@Stevoisiak Hi, we're going to add your filter here. Could you please review its name and description?

Alex-302 (approved):

AdamWr (approved):

piquark6046 (approved):

ntnguyen1234 (approved):

Stevoisiak:
> @Stevoisiak Hi, we're going to add your filter here. Could you please review its name and description?

Looks good to me!

Stevoisiak:
Out of curiosity, where do third-party filter lists in this repository get shown? Is there a list that gets served to users?

zloyden:
Ok, thank you for the review)

> Out of curiosity, where do third-party filter lists in this repository get shown? Is there a list that gets served to users?

You can take a look at them [here](https://github.com/AdguardTeam/FiltersRegistry/tree/master/filters/ThirdParty). Your filter will be there too.

Stevoisiak:
> You can take a look at them [here](https://github.com/AdguardTeam/FiltersRegistry/tree/master/filters/ThirdParty). Your filter will be there too.

I was more so asking how filter lists in this repo get presented to AdGuard users. Are they shown within the extension itself? The README also mentions that filters get rehosted and may be modified for compatibility. Do those modifications apply if a user directly adds the list via my repository?

Stevoisiak:
I see `ext_chromium_mv3` in the excluded platforms. The list should work in Chrome without the need for MV2. The only issues I encountered were in Opera specifically.

Alex-302:
> filters get rehosted and may be modified for compatibility

We can modify filters (removing problematic rules, if author cannot fix or if the problem related only to AdGuard. Also syntax can be converted. Validation may remove rules, considered as invalid). It is related only to files, hosted on our CDN.



> I see `ext_chromium_mv3` in the excluded platforms

@zloyden This filter is not large, less than 1500 rules. I think we can add it to mv3 extension.
The TPF was recently updated - CNAME email trackers were removed, resulting in a reduction of about 100,000 rules.

zloyden:
> Are they shown within the extension itself?

Yes, there is a list of filters, divided into groups, that users can use to choose which filters to enable or disable.

> Do those modifications apply if a user directly adds the list via my repository?

Well, yes, but it depends on the `trustLevel`, exclusions. In addition, what Alex-302 said above.

Stevoisiak:
> Yes, there is a list of filters, divided into groups, that users can use to choose which filters to enable or disable.

> This filter is not large, less than 1500 rules. I think we can add it to mv3 extension.

Ah, that explains it. I use AdGuard with Chrome and was confused why I didn't see filter lists from this repository. I didn't consider most lists would be hidden to stay within the MV3 rule limit, even if they were compatible.

filters/ThirdParty/filter_260_Stevos_AI_Blocklist/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+  "filterId": 260,
+  "name": "Stevo's AI Blocklist",
+  "description": "A filter list that hides specific website features that use Generative AI and content labeled as AI generated.",
```
abolshova-a11y:
```suggestion
  "description": "A filter list that hides website features using generative AI and AI-generated content.",
```

abolshova-a11y (changes requested):

Stevoisiak (changes requested):
I strongly prefer "features that use generative AI" over "features using generative AI" for the description. The latter makes it sound like the filter list itself uses AI.

abolshova-a11y:
> I strongly prefer "features that use generative AI" over "features using generative AI" for the description. The latter makes it sound like the filter list itself uses AI.

@Stevoisiak I don't think it comes off like that unless you're _really_ determined to connect it to a verb, but that's hardly natural sounding to me. In any case, I wanted to avoid the repetitive 'that', but if you feel like it doesn't work at all I can add it back.

Stevoisiak:
@abolshova-a11y I'd be ok with an alternate term like "features which use generative ai" to avoid the repetition.

abolshova-a11y:
@Stevoisiak ok fair enough, let's go with that

filters/ThirdParty/filter_260_Stevos_AI_Blocklist/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+  "filterId": 260,
+  "name": "Stevo's AI Blocklist",
+  "description": "A filter list that hides website features using generative AI and AI-generated content.",
```
abolshova-a11y:
```suggestion
  "description": "A filter list that hides website features which use generative AI and AI-generated content.",
```

abolshova-a11y (approved):

abolshova-a11y (approved):

PR #1209, opened 2026-07-07
Shorter description for Stevo's AI Blocklist

Stevoisiak:
No need to mention it being a filter list since every option in filters tab is a filter list. Better matches style of other annoyance filters.

PR #1217, opened 2026-08-11
Shorter description for Stevo's AI Blocklist

mu-hun:
Regarding the previous PR #1209,

> @stevoisiak: No need to mention it being a filter list since every option in the filters tab is a filter list. This better matches the style of other annoyance filters.

Sorry, I forgot to update the descriptions for other languages. I'll make sure to ask next time.

### DIFF
--- a/filters/ThirdParty/filter_260_Stevos_AI_Blocklist/metadata.json
+++ b/filters/ThirdParty/filter_260_Stevos_AI_Blocklist/metadata.json
@@ -1,7 +1,7 @@
 {
   "filterId": 260,
   "name": "Stevo's AI Blocklist",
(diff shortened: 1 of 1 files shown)


zloyden (approved):

abolshova-a11y (commented):
it's a pretty small change that doesn't change the meaning and that would require us to change translations in over 20 languages, i don't think it's worth it for just one filter...

zloyden:
@Stevoisiak [FYI](https://github.com/AdguardTeam/FiltersRegistry/pull/1217#pullrequestreview-4904187718).
@abolshova-a11y thanks, I also agree with you.

PR #1225, opened 2026-08-13
Replace List-KR filters list with "filterslist-KO Classic filter list"

piquark6046:


piquark6046:
Updated: Replacing List-KR filters list with filterslist-KO filters list entirely and dropping List-KR filters list as per discussion in AdGuard internal Slack.

zloyden (approved):

abolshova-a11y (approved):

abolshova-a11y (approved):

PR #1212, opened 2026-07-09
feat: add new npm script to compare the build output against master

mu-hun:
Turns the manual copy-paste workflow from `DEVELOPMENT.md` into a single interactive command. Please see details in the below issue.

> This PR is related to #1211 issue.

## Note

- About extra commits starting with `fix:`, which were not supported in the previous script.

Issue #1211, opened 2026-07-07
Create a regression test script to compare build results against master


mu-hun:
A workflow for comparing compiled rule output between a branch and master (worktrees, parallel builds, diff report).

## Problem

In the #1186 PR's instruction, it's currently a copy-paste shell snippet, which means:

- No progress feedback while the two builds run as background child processes
- No way to configure build behavior without editing the snippet inline
- Easy to typo or skip a step when pasting manually
- Cleanup step is manual and easy to forget

> Please check [Typical workflow — comparing build results against master in DEVELOPMENT.md](https://github.com/AdguardTeam/FiltersRegistry/blob/2bc697b1a5bb708407a51741179510b890f44ea2/DEVELOPMENT.md#typical-workflow--comparing-build-results-against-master)

## Proposal

Turn the workflow into a standalone script exposed via one npm alias, e.g.:

```bash
yarn compare-build-output
```

See the implemented design – https://github.com/AdguardTeam/FiltersRegistry/issues/1211#issuecomment-4912220425 below for the final shape — review feedback expanded it beyond the original requirements list, most notably with a shared optimization-stats cache between the two builds, a run lock, and signal handling.

## Note

Two behaviors from the manual doc are deliberately **not** carried over,
per discussion:

- The doc's cleanup step runs `git reset --hard` in the main repo as a blanket safety net. Baking an unconditional hard-reset into an automated script is a foot-gun (it would silently discard real uncommitted work if run from a dirty branch). The new script will not touch the main repo's tracked state at all — it only removes the worktrees and the `platforms_*_build/` dirs it created.
- No non-interactive/CI fallback for now (backlogged) — the script always prompts, matching local/manual usage. CI usage is a future concern.

## Requirements (original)

- **Single entry point**: one yarn script, no manual multi-step copy-paste
- **Structured report**: keep the existing pass/fail comparison output format
- **Progress indication**: show a loading spinner per build (master/changed), since both run as parallel child processes and currently give no feedback until done
- Interactive options — prompt for:
  - Branch selection: list local branches, defaulting to the currently checked-out branch.
  - Build mode: `generate-cache` + `build:local` vs. plain build
  - Whether to keep or clean up build artifacts (`platforms_*_build/` dirs, worktrees) after the run

mu-hun:
I need to check that the new npm script name should matches the existing `validate:*` conventions.

mu-hun:
## Design

```bash
yarn compare-build-output
[...]

PR #1226, opened 2026-08-14
Strips `version` and `timeUpdated` fields from filters.json and filters.js

### LINKED ISSUE #1208: Add strip generated metadata from filters.js, filters.json
To avoid irrelevant diffs when comparing comlided filters, following generated meta should de removed:

- `filters.js`
  - `timeUpdated`
  - `version`
- `filters.json`
  - `timeUpdated`
  - `version`

mu-hun:
Extends `--strip-generated-meta` (previously `.txt` only) to also drop these noisy fields from compiled metadata files. Closes #1208.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -86,7 +86,7 @@ for all supported AdGuard products.
 | `yarn build:patches` | Build incremental patches |
 | `yarn generate-cache` | Generate cached `filter.txt` from templates (`tsx scripts/build/build.js --generate-cache`) |
 | `yarn download-stats` | Download per-filter `stats.json` to `temp/optimization/stats/` |
-| `yarn strip-generated-meta` | Strip generated meta lines from platform filter files |
+| `yarn strip-generated-meta` | Strip meta lines from `.txt` files + `version`/`timeUpdated` from `filters.{json,js}` |
 | `yarn test` | Run unit tests (`vitest run`) |
 | `yarn lint` | Run all linters (code + types + markdown) |
 | `yarn lint:code` | ESLint check (`eslint . --ext .js,.ts`) |
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -179,6 +179,21 @@ General code style guidelines for JavaScript are available via link:
 - All other style rules (indentation, line length, Airbnb conventions, etc.)
   are enforced by `.eslintrc.cjs`. Run `yarn lint` to check.
 
+### Runtime data validation (TypeScript/JavaScript)
+
+Before adding any runtime check, decide whether the data is trusted or external. Data the
+system itself produced — own build artifacts, generated JSON, config files the project owns —
+is trusted: JSON.parse + as cast and move on. If such data is malformed, the code SHOULD fail
+loudly; never silently skip, default, or return-early on an unexpected shape.
+
+Genuinely external input (user input, network responses, third-party files) gets validated
+ONCE at the boundary — with a schema library (zod/valibot) when the project already has one —
+and the parsed type is trusted downstream. Do not re-validate past the boundary.
+
+Never write hand-rolled type guards (`isRecord`, `typeof x === 'object' && x !== null && !Array.isArray(x)` chains).
+A guard justified only by a hypothetical "what if the data is corrupt" is a smell:
+answer whether that corruption is real for THIS data source before writing anything.
+
 ### Testing
 
 - Test framework: Vitest.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -86,7 +86,7 @@ for all supported AdGuard products.
 | `yarn build:patches` | Build incremental patches |
 | `yarn generate-cache` | Generate cached `filter.txt` from templates (`tsx scripts/build/build.js --generate-cache`) |
 | `yarn download-stats` | Download per-filter `stats.json` to `temp/optimization/stats/` |
-| `yarn strip-generated-meta` | Strip meta lines from `.txt` files + `version`/`timeUpdated` from `filters.{json,js}` |
+| `yarn strip-generated-meta <dir>` | Strip generated meta from `.txt`/`filters.json`/`filters.js` in `<dir>` in place |
 | `yarn test` | Run unit tests (`vitest run`) |
 | `yarn lint` | Run all linters (code + types + markdown) |
 | `yarn lint:code` | ESLint check (`eslint . --ext .js,.ts`) |
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -79,6 +79,28 @@ The `yarn build` command (and `yarn build:local`) also accepts:
   files in `platforms/` and `temp/platforms/`, and remove the `version`/`timeUpdated`
   fields from every entry in `filters.json`/`filters.js`. Useful when comparing outputs between builds.
 
+### Stripping Generated Metadata
+
+`--strip-generated-meta` runs this pass as part of a build. To run the same pass
+on an existing build-output directory without recompiling, use the standalone
+command:
+
+```bash
+yarn strip-generated-meta <dir>
+```
+
+It rewrites, in place, every `.txt` filter file and every `filters.json` /
+`filters.js` under `<dir>` — dropping the generated meta lines and the
+`version` / `timeUpdated` fields, exactly as the build flag does.
+
+The directory argument is **required**. The command mutates files in place, and
+defaulting to the tracked `platforms/` tree would make it easy to strip the
+committed metadata the products read. Point it at a throwaway copy instead:
+
+```bash
+yarn strip-generated-meta temp/platforms
+```
+
 ### Generating Filter Cache
 
 To update the cached `filter.txt` files in `filters/`, used for
@@ -174,7 +196,9 @@ The following flags can be used with `yarn build` and `yarn build:local`:
   and not in `--skip`.
 - `--report=` — custom report file name (e.g., `--report='report-adguard.txt'`)
 - `--no-patches-prepare` — skip copying `platforms/` to `temp/platforms/`
-- `--strip-generated-meta` — remove volatile metadata lines from built files
+- `--strip-generated-meta` — remove generated meta lines from `.txt` files and
+  `version`/`timeUpdated` from `filters.json`/`filters.js`
+  (see [Stripping Generated Metadata](#stripping-generated-metadata))
 - `--use-cache` — build from cached `filter.txt` (same as `yarn build:local`)
 - `--generate-cache` — compile filters to update the `filter.txt` cache only
 - `--download-stats` — download each filter's `stats.json` without recompiling `filter.txt`.
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -76,7 +76,8 @@ The `yarn build` command (and `yarn build:local`) also accepts:
   (`yarn build:patches`) is not needed afterwards.
 - `--strip-generated-meta` — after compilation, remove generated meta lines
   (`! Checksum`, `! Diff-Path`, `! TimeUpdated`, `! Version`) from all filter
-  files in `platforms/` and `temp/platforms/`. Useful when comparing outputs between builds.
+  files in `platforms/` and `temp/platforms/`, and remove the `version`/`timeUpdated`
+  fields from every entry in `filters.json`/`filters.js`. Useful when comparing outputs between builds.
 
 ### Generating Filter Cache
 
--- a/scripts/build/__tests__/strip-generated-meta.test.ts
+++ b/scripts/build/__tests__/strip-generated-meta.test.ts
@@ -43,6 +43,33 @@ describe('stripGeneratedMetaFromDir', () => {
         await fs.rm(testDir, { recursive: true, force: true });
     });
 
+    /**
+     * Creates `<testDir>/<dirName>/platforms/cli/` and writes `content` verbatim
+     * to each of `fileNames` inside it.
+     *
+     * @param dirName - Unique sub-directory name isolating this case's tree.
+     * @param content - Exact bytes to write to every file.
+     * @param fileNames - Metadata file names to create (default: `filters.json`).
+     * @returns The `platforms` dir to pass to `stripGeneratedMetaFromDir`, plus the written file paths.
+     */
+    const writeMetadataFiles = async (
+        dirName: string,
+        content: string,
+        fileNames: string[] = ['filters.json'],
+    ): Promise<{ platformsDir: string; filePaths: string[] }> => {
+        const platformsDir = path.join(testDir, dirName, 'platforms');
+        const platformDir = path.join(platformsDir, 'cli');
+        await fs.mkdir(platformDir, { recursive: true });
+
+        const filePaths = await Promise.all(fileNames.map(async (name) => {
+            const filePath = path.join(platformDir, name);
+            await fs.writeFile(filePath, content, 'utf8');
+            return filePath;
+        }));
+
+        return { platformsDir, filePaths };
+    };
+
     it('strips generated metadata lines from .txt files', async () => {
         const modified = await stripGeneratedMetaFromDir(testDir);
 
@@ -119,9 +146,6 @@ describe('stripGeneratedMetaFromDir', () => {
     });
 
     it('strips version/timeUpdated fields from filters.json and filters.js', async () => {
-        const platformDir = path.join(testDir, 'json_root', 'platforms', 'cli');
-        await fs.mkdir(platformDir, { recursive: true });
-
         const metadata = {
             groups: [],
             tags: [],
@@ -137,21 +161,20 @@ describe('stripGeneratedMetaFromDir', () => {
             ],
         };
 
-        const jsonFilePath = path.join(platformDir, 'filters.json');
-        const jsFilePath = path.join(platformDir, 'filters.js');
         const content = JSON.stringify(metadata, null, '\t');
+        const { platformsDir, filePaths } = await writeMetadataFiles(
+            'json_root',
+            content,
+            ['filters.json', 'filters.js'],
+        );
 
-        await fs.writeFile(jsonFilePath, content, 'utf8');
-        await fs.writeFile(jsFilePath, content, 'utf8');
-
-        const modified = await stripGeneratedMetaFromDir(path.join(testDir, 'json_root', 'platforms'));
+        const modified = await stripGeneratedMetaFromDir(platformsDir);
 
         expect(modified).toBe(2);
 
-        const jsonContent = JSON.parse(await fs.readFile(jsonFilePath, 'utf8'));
-        const jsContent = JSON.parse(await fs.readFile(jsFilePath, 'utf8'));
+        const parsed = await Promise.all(filePaths.map(async (p) => JSON.parse(await fs.readFile(p, 'utf8'))));
 
-        [jsonContent, jsContent].forEach((data) => {
+        parsed.forEach((data) => {
             expect(data.filters[0]).not.toHaveProperty('version');
             expect(data.filters[0]).not.toHaveProperty('timeUpdated');
             expect(data.filters[0].filterId).toBe(1);
@@ -175,15 +198,10 @@ describe('stripGeneratedMetaFromDir', () => {
         ];
 
         await Promise.all(cases.map(async ({ dirName, hasTrailingNewline }) => {
-            const platformsDir = path.join(testDir, dirName, 'platforms');
-            const platformDir = path.join(platformsDir, 'cli');
-            await fs.mkdir(platformDir, { recursive: true });
-
             const serialized = JSON.stringify(metadata, null, '\t');
             const content = hasTrailingNewline ? `${serialized}\n` : serialized;
 
-            const jsonFilePath = path.join(platformDir, 'filters.json');
-            await fs.writeFile(jsonFilePath, content, 'utf8');
+            const { platformsDir, filePaths: [jsonFilePath] } = await writeMetadataFiles(dirName, content);
 
             await stripGeneratedMetaFromDir(platformsDir);
 
@@ -193,27 +211,22 @@ describe('stripGeneratedMetaFromDir', () => {
     });
 
     it('returns 0 for filters.json/filters.js with no version/timeUpdated fields', async () => {
-        const platformDir = path.join(testDir, 'json_root_clean', 'platforms', 'cli');
-        await fs.mkdir(platformDir, { recursive: true });
-
         const metadata = { groups: [], tags: [], filters: [{ filterId: 1, name: 'Test filter' }] };
-        const jsonFilePath = path.join(platformDir, 'filters.json');
-        await fs.writeFile(jsonFilePath, JSON.stringify(metadata, null, '\t'), 'utf8');
+        const { platformsDir } = await writeMetadataFiles('json_root_clean', JSON.stringify(metadata, null, '\t'));
 
-        const modified = await stripGeneratedMetaFromDir(path.join(testDir, 'json_root_clean', 'platforms'));
+        const modified = await stripGeneratedMetaFromDir(platformsDir);
         expect(modified).toBe(0);
     });
 
     it('throws and leaves the file untouched when filters.json contains invalid JSON', async () => {
-        const platformDir = path.join(testDir, 'json_root_invalid', 'platforms', 'cli');
-        await fs.mkdir(platformDir, { recursive: true });
-
-        const jsonFilePath = path.join(platformDir, 'filters.json');
         const invalidContent = '{ "filters": [ invalid json here';
-        await fs.writeFile(jsonFilePath, invalidContent, 'utf8');
+        const {
+            platformsDir,
+            filePaths: [jsonFilePath],
+        } = await writeMetadataFiles('json_root_invalid', invalidContent);
 
         await expect(
-            stripGeneratedMetaFromDir(path.join(testDir, 'json_root_invalid', 'platforms')),
+            stripGeneratedMetaFromDir(platformsDir),
         ).rejects.toThrow(/Failed to parse metadata file/);
 
         const contentAfter = await fs.readFile(jsonFilePath, 'utf8');
@@ -228,14 +241,13 @@ describe('stripGeneratedMetaFromDir', () => {
         ];
 
         await Promise.all(cases.map(async ({ dirName, metadata }) => {
-            const platformDir = path.join(testDir, dirName, 'platforms', 'cli');
-            await fs.mkdir(platformDir, { recursive: true });
-
-            const jsonFilePath = path.join(platformDir, 'filters.json');
             const content = JSON.stringify(metadata, null, '\t');
-            await fs.writeFile(jsonFilePath, content, 'utf8');
+            const {
+                platformsDir,
+                filePaths: [jsonFilePath],
+            } = await writeMetadataFiles(dirName, content);
 
-            const modified = await stripGeneratedMetaFromDir(path.join(testDir, dirName, 'platforms'));
+            const modified = await stripGeneratedMetaFromDir(platformsDir);
             expect(modified).toBe(0);
 
             const contentAfter = await fs.readFile(jsonFilePath, 'utf8');
@@ -244,15 +256,11 @@ describe('stripGeneratedMetaFromDir', () => {
     });
 
     it('returns 0 and leaves the file untouched when filters array entries are null or arrays', async () => {
-        const platformDir = path.join(testDir, 'json_root_bad_entries', 'platforms', 'cli');
-        await fs.mkdir(platformDir, { recursive: true });
-
         const metadata = { groups: [], tags: [], filters: [null, ['nested', 'array']] };
-        const jsonFilePath = path.join(platformDir, 'filters.json');
         const content = JSON.stringify(metadata, null, '\t');
-        await fs.writeFile(jsonFilePath, content, 'utf8');
+        const { platformsDir, filePaths: [jsonFilePath] } = await writeMetadataFiles('json_root_bad_entries', content);
 
-        const modified = await stripGeneratedMetaFromDir(path.join(testDir, 'json_root_bad_entries', 'platforms'));
+        const modified = await stripGeneratedMetaFromDir(platformsDir);
         expect(modified).toBe(0);
 
         const contentAfter = await fs.readFile(jsonFilePath, 'utf8');
@@ -260,15 +268,14 @@ describe('stripGeneratedMetaFromDir', () => {
     });
 
     it('returns 0 and leaves the file untouched when the filters array is empty', async () => {
-        const platformDir = path.join(testDir, 'json_root_empty_filters', 'platforms', 'cli');
-        await fs.mkdir(platformDir, { recursive: true });
-
         const metadata = { groups: [], tags: [], filters: [] };
-        const jsonFilePath = path.join(platformDir, 'filters.json');
         const content = JSON.stringify(metadata, null, '\t');
-        await fs.writeFile(jsonFilePath, content, 'utf8');
+        const {
+            platformsDir,
+            filePaths: [jsonFilePath],
+        } = await writeMetadataFiles('json_root_empty_filters', content);
 
-        const modified = await stripGeneratedMetaFromDir(path.join(testDir, 'json_root_empty_filters', 'platforms'));
+        const modified = await stripGeneratedMetaFromDir(platformsDir);
         expect(modified).toBe(0);
 
         const contentAfter = await fs.readFile(jsonFilePath, 'utf8');
--- a/scripts/build/__tests__/strip-generated-meta.test.ts
+++ b/scripts/build/__tests__/strip-generated-meta.test.ts
@@ -160,6 +160,38 @@ describe('stripGeneratedMetaFromDir', () => {
         });
     });
 
+    it('preserves tab indentation and trailing-newline from filters.json and filters.js', async () => {
+        const preservedFilter = { filterId: 1, name: 'Test filter' };
+        const metadata = {
+            groups: [],
+            tags: [],
+            filters: [{ ...preservedFilter, version: '2.1.8.0', timeUpdated: '2026-08-10T18:21:50+0000' }],
+        };
+        const expectedSerialized = JSON.stringify({ ...metadata, filters: [preservedFilter] }, null, '\t');
+
+        const cases = [
+            { dirName: 'json_root_raw_no_newline', hasTrailingNewline: false },
+            { dirName: 'json_root_raw_with_newline', hasTrailingNewline: true },
+        ];
+
+        await Promise.all(cases.map(async ({ dirName, hasTrailingNewline }) => {
+            const platformsDir = path.join(testDir, dirName, 'platforms');
+            const platformDir = path.join(platformsDir, 'cli');
+            await fs.mkdir(platformDir, { recursive: true });
+
+            const serialized = JSON.stringify(metadata, null, '\t');
+            const content = hasTrailingNewline ? `${serialized}\n` : serialized;
+
+            const jsonFilePath = path.join(platformDir, 'filters.json');
+            await fs.writeFile(jsonFilePath, content, 'utf8');
+
+            await stripGeneratedMetaFromDir(platformsDir);
+
+            const rawContent = await fs.readFile(jsonFilePath, 'utf8');
+            expect(rawContent).toBe(hasTrailingNewline ? `${expectedSerialized}\n` : expectedSerialized);
+        }));
+    });
+
     it('returns 0 for filters.json/filters.js with no version/timeUpdated fields', async () => {
         const platformDir = path.join(testDir, 'json_root_clean', 'platforms', 'cli');
         await fs.mkdir(platformDir, { recursive: true });
--- a/scripts/build/__tests__/strip-generated-meta.test.ts
+++ b/scripts/build/__tests__/strip-generated-meta.test.ts
@@ -187,4 +187,59 @@ describe('stripGeneratedMetaFromDir', () => {
         const contentAfter = await fs.readFile(jsonFile, 'utf8');
         expect(contentAfter).toBe(invalidContent);
     });
+
+    it('returns 0 and leaves the file untouched when the filters field is missing, null, or not an array', async () => {
+        const cases = [
+            { dirName: 'json_root_missing_filters', metadata: { groups: [], tags: [] } },
+            { dirName: 'json_root_null_filters', metadata: { groups: [], tags: [], filters: null } },
+            { dirName: 'json_root_nonarray_filters', metadata: { groups: [], tags: [], filters: 'not-an-array' } },
+        ];
+
+        await Promise.all(cases.map(async ({ dirName, metadata }) => {
+            const platformDir = path.join(testDir, dirName, 'platforms', 'cli');
+            await fs.mkdir(platformDir, { recursive: true });
+
+            const jsonFile = path.join(platformDir, 'filters.json');
+            const content = JSON.stringify(metadata, null, '\t');
+            await fs.writeFile(jsonFile, content, 'utf8');
+
+            const modified = await stripGeneratedMetaFromDir(path.join(testDir, dirName, 'platforms'));
+            expect(modified).toBe(0);
+
+            const contentAfter = await fs.readFile(jsonFile, 'utf8');
+            expect(contentAfter).toBe(content);
+        }));
+    });
+
+    it('returns 0 and leaves the file untouched when filters array entries are null or arrays', async () => {
+        const platformDir = path.join(testDir, 'json_root_bad_entries', 'platforms', 'cli');
+        await fs.mkdir(platformDir, { recursive: true });
+
+        const metadata = { groups: [], tags: [], filters: [null, ['nested', 'array']] };
+        const jsonFile = path.join(platformDir, 'filters.json');
+        const content = JSON.stringify(metadata, null, '\t');
+        await fs.writeFile(jsonFile, content, 'utf8');
+
+        const modified = await stripGeneratedMetaFromDir(path.join(testDir, 'json_root_bad_entries', 'platforms'));
+        expect(modified).toBe(0);
+
+        const contentAfter = await fs.readFile(jsonFile, 'utf8');
+        expect(contentAfter).toBe(content);
+    });
+
+    it('returns 0 and leaves the file untouched when the filters array is empty', async () => {
+        const platformDir = path.join(testDir, 'json_root_empty_filters', 'platforms', 'cli');
+        await fs.mkdir(platformDir, { recursive: true });
+
+        const metadata = { groups: [], tags: [], filters: [] };
+        const jsonFile = path.join(platformDir, 'filters.json');
+        const content = JSON.stringify(metadata, null, '\t');
+        await fs.writeFile(jsonFile, content, 'utf8');
+
+        const modified = await stripGeneratedMetaFromDir(path.join(testDir, 'json_root_empty_filters', 'platforms'));
+        expect(modified).toBe(0);
+
+        const contentAfter = await fs.readFile(jsonFile, 'utf8');
+        expect(contentAfter).toBe(content);
+    });
 });
--- a/scripts/build/__tests__/strip-generated-meta.test.ts
+++ b/scripts/build/__tests__/strip-generated-meta.test.ts
@@ -117,4 +117,58 @@ describe('stripGeneratedMetaFromDir', () => {
         expect(oldContent.includes('! Version:')).toBe(false);
         expect(oldContent.includes('||example.com^')).toBe(true);
     });
+
+    it('strips version/timeUpdated fields from filters.json and filters.js', async () => {
+        const platformDir = path.join(testDir, 'json_root', 'platforms', 'cli');
+        await fs.mkdir(platformDir, { recursive: true });
+
+        const metadata = {
+            groups: [],
+            tags: [],
+            filters: [
+                {
+                    filterId: 1,
+                    name: 'Test filter',
+                    timeAdded: '2014-06-30T07:56:55+0000',
+                    version: '2.1.8.0',
+                    timeUpdated: '2026-08-10T18:21:50+0000',
+                    deprecated: false,
+                },
+            ],
+        };
+
+        const jsonFile = path.join(platformDir, 'filters.json');
+        const jsFile = path.join(platformDir, 'filters.js');
+        const content = JSON.stringify(metadata, null, '\t');
+
+        await fs.writeFile(jsonFile, content, 'utf8');
+        await fs.writeFile(jsFile, content, 'utf8');
+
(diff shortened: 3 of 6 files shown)


scripts/build/strip-generated-meta.ts
```diff
@@ -64,11 +105,12 @@ const walkAndStrip = async (dir: string, rootDir: string): Promise<number> => {
...
         if (!entry.isDirectory()) {
-            // Only process .txt files that are directly inside a `filters/` dir,
-            // which is guaranteed by the caller when dir itself is a filters dir.
             if (entry.name.endsWith(TXT_FILE_EXTENSION)) {
                 return (await stripGeneratedMeta(fullPath)) ? 1 : 0;
             }
+            if (METADATA_FILE_NAMES.includes(entry.name)) {
+                return (await stripMetaFromMetadataFile(fullPath)) ? 1 : 0;
+            }
             return 0;
```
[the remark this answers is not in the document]
mu-hun: This review doesn't seem important or has a lower priority because all the compiled `.txt` filters are under the `filters/` directories, except for the `local_script_rules.txt` file located under each platform directory.

scripts/build/strip-generated-meta.ts
```diff
@@ -27,6 +34,57 @@ const isGeneratedMetaLine = (line: string): boolean => {
...
+ *
+ * @param filePath - Absolute path to the filters.json or filters.js file.
+ * @returns True if the file was modified, false otherwise.
+ */
+const stripMetaFromMetadataFile = async (filePath: string): Promise<boolean> => {
+    const content = await fs.readFile(filePath, 'utf8');
+    let data: { filters?: unknown };
+
+    try {
+        data = JSON.parse(content) as { filters?: unknown };
```
105th: The parsed value is cast directly to `{ filters?: unknown }` without a runtime check that it is actually an object. If a metadata file contains the literal `null` (or a corrupt build writes one), `JSON.parse` returns `null`, and the subsequent `Array.isArray(data.filters)` access throws `TypeError: Cannot read properties of null (reading 'filters')`. This crashes the whole build with a confusing error instead of returning `false` via the intended guard.

Recommendation: verify the parsed value is a non-null, non-array object before accessing `.filters`:

```ts
const parsed = JSON.parse(content) as unknown;
if (typeof parsed !== 'object' || parsed === null || Array.isArray(parsed)) {
    return false;
}
const data = parsed as { filters?: unknown };
```

scripts/build/strip-generated-meta.ts
```diff
@@ -27,6 +34,57 @@ const isGeneratedMetaLine = (line: string): boolean => {
...
+    } catch (error) {
+        const reason = error instanceof Error ? ` ${error.message}` : '';
+        throw new Error(
+            `Failed to parse metadata file at ${filePath}. The file must contain valid JSON.${reason}`,
+        );
```
105th, line 58: Now that `METADATA_FILE_NAMES` includes `filters.js`, this function unconditionally requires the target to be strict JSON. The current `platforms/*/filters.js` files happen to be valid JSON, but the `.js` extension only guarantees a JavaScript file: any platform that emits a real JS module (`module.exports = {...}`, comments, single quotes, or trailing commas) will make `JSON.parse` throw, and the thrown error propagates through `walkAndStrip` → `Promise.all`, aborting the entire multi-platform build even though the goal here is best-effort diff-noise reduction. The message 'The file must contain valid JSON.' is also misleading for a `.js` file.

Recommendation: prefer a targeted field-strip approach for `.js`, or at least make the parse failure non-fatal (log and skip) with a file-type-aware message.

scripts/build/strip-generated-meta.ts
```diff
@@ -27,6 +34,57 @@ const isGeneratedMetaLine = (line: string): boolean => {
     return GENERATED_META_PREFIXES.some((prefix) => line.startsWith(prefix));
 };
 
+/**
+ * Checks whether an array contains only non-null objects (i.e. records).
+ * @param value - The array to check.
+ * @returns True if every item in the array is a non-null object.
+ */
+const isRecordArray = (value: unknown[]): value is Record<string, unknown>[] => {
+    return value.every((item) => typeof item === 'object' && item !== null);
```
105th: This type guard is unsound: `typeof item === 'object'` is also true for arrays (`typeof [] === 'object'`), so an array entry inside the `filters` array would pass the guard and be typed as `Record<string, unknown>`. The subsequent `field in filter` / `delete filter[field]` operations would then run against an array instead of an object.

Fix:

```ts
return value.every((item) => typeof item === 'object' && item !== null && !Array.isArray(item));
```

scripts/build/strip-generated-meta.ts
```diff
@@ -27,6 +34,57 @@ const isGeneratedMetaLine = (line: string): boolean => {
...
+                modified = true;
+            }
+        });
+    });
+
+    if (!modified) {
+        return false;
+    }
+
+    await fs.writeFile(filePath, JSON.stringify(data, null, '\t'), 'utf8');
```
105th: The in-place edit re-serializes the entire file with `JSON.stringify(data, null, '\t')`, so the write is a full reformat rather than a surgical field removal. `JSON.stringify` drops any trailing newline the compiled file may have and normalizes indentation to tabs — exactly the kind of formatting-only change issue #1208 is trying to keep out of the branch-vs-master comparison.

Recommendation: preserve the original trailing newline (and ideally the original indentation) so the only diff introduced is the removed `version`/`timeUpdated` fields:

```ts
const serialized = JSON.stringify(data, null, '\t');
await fs.writeFile(filePath, content.endsWith('\n') ? `${serialized}\n` : serialized, 'utf8');
```

scripts/compare-build-output-against-master.sh
```diff
@@ -223,7 +223,7 @@ generate_report() {
   [ -n "$diff_list" ] && printf "$diff_list\n"
   echo ""
   echo "${C_BOLD}--- Metadata Files (informational) ---${C_RESET}"
-  echo "filters.json/filters.js diffs: $meta_diffs  (version counter noise, not a regression)"
+  echo "filters.json/filters.js diffs: $meta_diffs (a nonzero count here signals a metadata change)"
```
105th: The relabeled message claims a nonzero count always signals a true metadata change, but that holds only once master is rebuilt with this PR's strip logic. In this script the master worktree is built with its own pre-PR `build.js`/`strip-generated-meta.ts`, which strips only `.txt` files and still writes `version`/`timeUpdated` into `filters.json`/`filters.js`. So during this PR's own validation run, every `filters.json`/`filters.js` will differ purely from the strip asymmetry (master retains the fields, the feature branch deletes them) — producing a nonzero meta count that is exactly the version noise issue #1208 aims to eliminate, not a metadata change.

Recommendation: cover this transitional case, e.g. note that a nonzero count may be `version`/`timeUpdated` noise until master includes the metadata strip.

105th (commented):
## AI Code Review (`open-code-review`)

Findings from an automated review of this PR (extends `--strip-generated-meta` to remove `version`/`timeUpdated` from `filters.json`/`filters.js`, closing #1208).

Summary: 5 findings — 4 medium, 1 low. No blocking (high) issues found. Main themes: (1) robustness of the new `stripMetaFromMetadataFile` against non-JSON/`null` inputs, (2) avoiding the exact formatting noise this PR aims to eliminate, and (3) a documentation label in `compare-build-output-against-master.sh` that will mislead for this PR's own validation run.

Details in the inline comments below.

105th (approved):

scripts/compare-build-output-against-master.sh
mu-hun: As I understand it, the new specific content changes are only from `$CHANGED_BRANCH` and are not included in `$BASED_BRANCH`. Is my understanding correct? However, I've already added a step to sync the `filters/` directory using `git checkout $MASTER_SHA -- filters/`. https://github.com/AdguardTeam/FiltersRegistry/blob/b5b149149f1ad884c9c224b25ef67152dfddf51c/scripts/compare-build-output-against-master.sh#L385

This overwrites the changed worktree's `filters/` (including `revision.json`) with master's exact content, so both worktrees compile from identical source.

scripts/build/strip-generated-meta.ts
```diff
@@ -27,6 +34,57 @@ const isGeneratedMetaLine = (line: string): boolean => {
...
+ * @param filePath - Absolute path to the filters.json or filters.js file.
+ * @returns True if the file was modified, false otherwise.
+ */
+const stripMetaFromMetadataFile = async (filePath: string): Promise<boolean> => {
+    const content = await fs.readFile(filePath, 'utf8');
+    if (typeof content !== 'string' ) {
+        return false;
+    }
+
+    const parsed = JSON.parse(content) as unknown;
```
105th: `typeof content !== 'string'` can never be true here — `fs.readFile(path, 'utf8')` always resolves to a string, so this guard is dead code that only *looks* defensive. Meanwhile `JSON.parse(content)` is not wrapped in try/catch: a single malformed `filters.json`/`filters.js` found anywhere under the root throws a raw `SyntaxError`, propagates through the `Promise.all` in `walkAndStrip`, and aborts the entire `--strip-generated-meta` build step (`build.js` awaits `stripGeneratedMetaFromDir` with no error handling), and the error contains no file path.

This is also a regression introduced by this PR: commit `44cd3e971` previously threw a contextual `Failed to parse metadata file at <path>` error, and `539cf20e9` replaced it with this guard. Since the function deliberately returns `false` for every other unexpected shape, please keep that contract for invalid JSON as well:

```ts
let parsed: unknown;
try {
    parsed = JSON.parse(content);
} catch (error) {
    const reason = error instanceof Error ? ` ${error.message}` : '';
    throw new Error(
        `Failed to parse metadata file at ${filePath}. The file must contain valid JSON.${reason}`,
    );
}
```

scripts/build/__tests__/strip-generated-meta.test.ts
```diff
@@ -117,4 +117,58 @@ describe('stripGeneratedMetaFromDir', () => {
         expect(oldContent.includes('! Version:')).toBe(false);
         expect(oldContent.includes('||example.com^')).toBe(true);
     });
+
+    it('strips version/timeUpdated fields from filters.json and filters.js', async () => {
```
105th, line 148: The new tests only cover the happy path and the already-clean path. The graceful-degradation branches aren't covered: invalid JSON (which currently throws — see `strip-generated-meta.ts:59`), a missing/`null`/non-array `filters` field, entries that are `null`/arrays, and an empty `filters` array. Please add cases pinning that these inputs are safely skipped without crashing or touching the file — especially invalid JSON, which would abort the whole build today.

scripts/build/__tests__/strip-generated-meta.test.ts
```diff
@@ -117,4 +117,58 @@ describe('stripGeneratedMetaFromDir', () => {
...
+        const content = JSON.stringify(metadata, null, '\t');
+
+        await fs.writeFile(jsonFile, content, 'utf8');
+        await fs.writeFile(jsFile, content, 'utf8');
+
+        const modified = await stripGeneratedMetaFromDir(path.join(testDir, 'json_root', 'platforms'));
+
+        expect(modified).toBe(2);
+
+        const jsonContent = JSON.parse(await fs.readFile(jsonFile, 'utf8'));
```
105th: Assertions re-read the output with `JSON.parse`, which is whitespace-insensitive. A regression that switches serialization from tabs to spaces, or that drops/appends the trailing newline (the implementation appends `\n` only when the source file had one), would pass CI while producing noisy diffs in `compare-build-output-against-master.sh` — the exact thing this PR exists to prevent. Consider also asserting the raw file content (tab indentation, trailing-newline preservation).

105th (changes requested):
Reviewed the full diff (4 files; `DEVELOPMENT.md` skipped as docs). The stripping logic itself looks correct — metadata-file matching, the record-array guard, and the modified-file counting all check out, and the tab-indented serialization matches what `@adguard/filters-compiler` writes (both files are plain JSON, no trailing newline), so stripping is self-consistent for the compare workflow.

One blocking concern (a dead guard + a lost try/catch that can abort the build with a file-path-less error) and two test-coverage gaps below.

Minor (non-blocking): in the "clean file" case (`strip-generated-meta.test.ts:171`) consider asserting the file content stays byte-identical to pin the deliberate do-not-rewrite behavior.

scripts/build/__tests__/strip-generated-meta.test.ts
mu-hun: Other review comments have been resolved. I will check-in this issue after dinner (after KST 6PM) 👀

105th (approved):
## Code Review Summary

Ran an independent AI code review (`ocr` / open-code-review) on the full diff (5 files, +245/−19).

**Verdict: No actionable issues found.** The implementation is solid.

### What this PR does
Extends `--strip-generated-meta` to also remove `version`/`timeUpdated` fields from `filters.json`/`filters.js` metadata files, in addition to the existing `.txt` meta-line stripping. Goal: clean, reproducible build-output diffs.

### Verification highlights
- **`filters.js` is strict JSON**: Confirmed against the actual committed `platforms/cli/filters.js` and `platforms/cli/filters.json` — both parse cleanly with `JSON.parse` (tab-indented, LF, no CRLF, contains real `version`/`timeUpdated` fields). The compiler emits strict JSON with a `.js` extension, not a JS module, so `JSON.parse` is safe.
- **Robustness guards are correct**: `isRecordArray` properly excludes `null` and arrays via `!Array.isArray(item)`; the parsed value is checked for non-null/non-array before accessing `.filters`; `try/catch` around `JSON.parse` rethrows with a contextual file-path message; missing/null/non-array/`filters` and bad entries return `0` gracefully.
- **Formatting preserved**: `JSON.stringify(data, null, '\t')` matches the compiler's own tab indentation, and the trailing-newline is preserved via `content.endsWith('\n')` — no spurious full-file reformatting.
- **Test coverage is thorough**: strips success, indentation/trailing-newline preservation, no-op on clean files, throw-on-invalid-JSON (file left untouched), missing/null/non-array `filters`, null/array entries, and empty `filters` array.

Approved.

mu-hun:
Thanks for your approval! I'd suggest just holding the current state until the parent PR is merged, rather than merging this into another PR. Since I'm following a stacked PR workflow (https://www.stacking.dev/), I'd like to keep each PR as a single commit when merging into master.

mu-hun:
Note on the `filters.json/filters.js` diffs: 32 count from `compare-build-output-against-master.sh`:

```
--- Metadata Files (informational) ---
filters.json/filters.js diffs: 32 (a nonzero count here signals a metadata change)
```

This is expected, not a regression. Master's `strip-generated-meta.ts` doesn't have `stripMetaFromMetadataFile` / `GENERATED_META_FIELDS` / `METADATA_FILE_NAMES` — it only strips generated lines from `.txt` filter files and never touches `filters.json/filters.js`. That capability is added in this PR, so master's build output still carries `version/timeUpdated`, while this branch's output has them stripped:

```diff
diff temp/platforms_master_build/mac/filters.json temp/platforms_changed_build/mac/filters.json
49,50d48
-                       "version": "2.1.8.22",
-                       "timeUpdated": "2026-08-21T19:13:03+0000",
64,65d61
```

Once this feature is merged to master, the metadata-diff count against master will drop to 0 for unrelated branches. Until then, any branch compared against current master will show a nonzero count here. That's a known baseline mismatch, not a signal to investigate.

zloyden (approved):

Alex-302:
Some of the phrases in the script interface are a little long.

> Found build output from a previous run (feature/#1208 vs master, plain) — generate the report from it now instead of rebuilding?

=> `Output directory is not empty ($CHANGED_BRANCH vs $BASED_BRANCH, $MODE_LABEL). Reuse it, or [delete and] rebuild?` Without `[delete and]` also clear I think.
Current default is No. I'd prefer to change to Yes, since we may not be sure exactly what is in the old files.

> Ignoring previous build output, rebuilding

old files removed or not? If removed => `Removing previous build output, rebuilding`

> Step 3/9: Cleanup preference
> Remove worktrees and platforms_*_build/ when done?

=> `Remove build output when done?`

> Step 4/9: Set up worktrees
> [master] Existing worktree found at C:/Work/AdGuard/Repos/AdGuard/Main/FiltersRegistry/temp/reg-master-build — reuse it instead of recreate from scratch?

Long line, more readable as two

=> `[master] Existing worktree found at C:/Work/AdGuard/Repos/AdGuard/Main/FiltersRegistry/temp/reg-master-build`
=> `Reuse it instead of rebuilding?`

mu-hun:
Since here is nighttime, I will apply your comment tomorrow morning. 💤

scripts/build/strip-generated-meta.ts
```diff
@@ -27,6 +34,63 @@ const isGeneratedMetaLine = (line: string): boolean => {
...
+ * @param value - The array to check.
+ * @returns True if every item in the array is a non-null object.
+ */
+const isRecordArray = (value: unknown[]): value is Record<string, unknown>[] => {
+    return value.every((item) => typeof item === 'object' && item !== null && !Array.isArray(item));
+};
+
+/**
+ * Strip generated version/timeUpdated fields from each entry of the `filters` array
+ * in a filters.json/filters.js metadata file, in-place.
```
slvvko, line 39:
```suggestion
 * in a filters.json/filters.js metadata file, in-place.
 *
 * https://github.com/AdguardTeam/FiltersRegistry/issues/1208
```

slvvko (approved):

Alex-302 (approved):

maximtop (commented):
The verdict change is the thing to settle — as written every comparison fails on metadata until this reaches master, which is the same asymmetry the open thread on this file raises. Rest is inline.

scripts/compare-build-output-against-master.sh
```diff
@@ -222,13 +222,19 @@ generate_report() {
...
-  echo "${C_BOLD}--- Metadata Files (informational) ---${C_RESET}"
-  echo "filters.json/filters.js diffs: $meta_diffs  (version counter noise, not a regression)"
+  echo "${C_BOLD}--- Metadata Files ---${C_RESET}"
+  echo "filters.json/filters.js diffs: $meta_diffs"
   echo ""
   echo "${C_BOLD}--- Verdict ---${C_RESET}"
-  [ "$rule_diffs" -eq 0 ] \
-    && echo "${C_GREEN}${CHECK} PASS${C_RESET} — no rule file diffs" \
-    || { echo "${C_RED}${CROSS} FAIL${C_RESET} — $rule_diffs rule file(s) differ"; return 1; }
+  if [ "$rule_diffs" -eq 0 ] && [ "$meta_diffs" -eq 0 ]; then
```
maximtop: As written this fails every run during the transition window: the master worktree builds with master's checkout, which has no `--strip-generated-meta`, so `filters.json`/`filters.js` always differ in `version`/`timeUpdated` (~32 metadata diffs → FAIL) until this lands on master. This is the same asymmetry the still-open thread on this file raises (#3813020249), and the author's reply there doesn't address the script side. Either ignore the volatile fields in the metadata comparison, or keep the metadata check informational until the flag exists on master.

scripts/build/strip-generated-meta.ts
```diff
@@ -48,7 +114,8 @@ const stripGeneratedMeta = async (filePath: string): Promise<boolean> => {
 
 /**
  * Single-pass recursive walk: strips metadata from .txt files inside `filters/`
```
maximtop, line 109: This says .txt files are stripped "inside `filters/` directories", but the non-directory branch below strips any `.txt` at any depth under the root. Either behavior could be right for the compare use case, but the doc and code should agree (this was also the point of the earlier open thread on this file).

scripts/build/strip-generated-meta.ts
```diff
@@ -64,11 +131,12 @@ const walkAndStrip = async (dir: string, rootDir: string): Promise<number> => {
         const fullPath = path.join(dir, entry.name);
 
         if (!entry.isDirectory()) {
-            // Only process .txt files that are directly inside a `filters/` dir,
-            // which is guaranteed by the caller when dir itself is a filters dir.
             if (entry.name.endsWith(TXT_FILE_EXTENSION)) {
                 return (await stripGeneratedMeta(fullPath)) ? 1 : 0;
             }
+            if (METADATA_FILE_NAMES.includes(entry.name)) {
```
maximtop, line 131: When a `filters.json`/`filters.js` gets stripped at a platform root there's no log line — only `filters/` directories get the aggregate "stripped metadata from N file(s)" message. Since stripping these files is the core of the feature, an occasional confirmation in the output would help verify it actually did something.

scripts/compare-build-output-against-master.sh
```diff
@@ -222,13 +222,19 @@ generate_report() {
...
+    echo "${C_GREEN}${CHECK} PASS${C_RESET} — no rule file or metadata diffs"
+  else
+    local fail_reasons=""
+    [ "$rule_diffs" -gt 0 ] && fail_reasons="$fail_reasons $rule_diffs rule file(s) differ;"
```
maximtop: `fail_reasons` appends `;` after every entry and nothing trims the last one, so the verdict prints with a dangling semicolon: `✖ FAIL — 3 rule file(s) differ; 32 metadata file(s) differ;`

scripts/build/__tests__/strip-generated-meta.test.ts
```diff
@@ -117,4 +117,161 @@ describe('stripGeneratedMetaFromDir', () => {
         expect(oldContent.includes('! Version:')).toBe(false);
         expect(oldContent.includes('||example.com^')).toBe(true);
     });
+
+    it('strips version/timeUpdated fields from filters.json and filters.js', async () => {
```
maximtop, line 148: The six metadata-file tests below repeat the same mkdir + writeFile scaffolding. A small `writeJson(platformDir, value)` helper would collapse most of it and keep the tests focused on their assertions.

scripts/build/strip-generated-meta.ts
```diff
@@ -27,6 +34,65 @@ const isGeneratedMetaLine = (line: string): boolean => {
     return GENERATED_META_PREFIXES.some((prefix) => line.startsWith(prefix));
 };
 
+/**
+ * Checks whether an array contains only non-null objects (i.e. records).
```
maximtop: Missing the blank line between the description and `@param` that the surrounding JSDoc blocks use.

maximtop:
One thing outside the diff: the AGENTS.md command-table row for `yarn strip-generated-meta` says "Strip generated meta lines from platform filter files", which no longer covers what the flag does after this PR — it also removes the `version`/`timeUpdated` fields from `filters.json`/`filters.js`. Worth updating the row here or in a follow-up.

slvvko (approved):

maximtop (approved):

slvvko (approved):

scripts/build/strip-generated-meta.ts
mu-hun: I've seen this too and tracked it in #1232. 

This PR doesn't need to stay open for it, as the problem exists independently of these changes. I've labeled the issue as low priority.

scripts/compare-build-output-against-master.sh
mu-hun: 👀 I'll leave the `filters.json/filters.js` metadata check as a non-blocking information alert untill this PR lands in master.

scripts/build/strip-generated-meta.ts
```diff
@@ -27,6 +34,66 @@ const isGeneratedMetaLine = (line: string): boolean => {
...
+ *
+ * @param value - The array to check.
+ * @returns True if every item in the array is a non-null object.
+ */
+const isRecordArray = (value: unknown[]): value is Record<string, unknown>[] => {
```
maximtop: Image

mu-hun: 😂 I'm wondering if we should drop this type guard, or maybe you have a better idea in mind?

maximtop: we don't need to validate the data generated in the previous step

maximtop: also, you can add rules to guide the llm like I mentioned here

https://performix.slack.com/archives/CVDTRTE91/p1788419998150279

maximtop (commented):
Main thing before merge: the metadata count now hard-fails every compare run until the strip lands on master — details in the reply on the open thread. The rest is inline; the printf change and the dropped Added-Files section are the two that actually need a decision, everything else is small cleanups.

scripts/compare-build-output-against-master.sh
```diff
@@ -371,22 +371,24 @@ generate_report() {
...
-    echo "filters.json/filters.js diffs: $meta_diffs  (version counter noise, not a regression)"
-    echo ""
-    echo "${C_BOLD}--- Verdict ---${C_RESET}"
-    [ "$rule_diffs" -eq 0 ] \
-        && echo "${C_GREEN}${CHECK} PASS${C_RESET} — no rule file diffs" \
-        || { echo "${C_RED}${CROSS} FAIL${C_RESET} — $rule_diffs rule file(s) differ"; return 1; }
+  echo "${C_BOLD}--- Rule Files ---${C_RESET}"
+  echo "Total .txt files compared: $total"
+  echo "Files with diffs:          $rule_diffs"
+  [ -n "$diff_list" ] && printf "$diff_list\n"
```
maximtop: `printf "$diff_list\n"` makes the accumulated paths the format string — a `%` in a filename prints an `invalid format character` error and truncates the entry, and backslashes get re-interpreted. `printf '%b\n' "$diff_list"` did the same `\n` expansion with the data as an argument. Let's keep the old form.

scripts/compare-build-output-against-master.sh
```diff
@@ -371,22 +371,24 @@ generate_report() {
         "$PLATFORMS_MASTER" "$PLATFORMS_CHANGED" \
         | grep -cE "filters\.(json|js)" || true)
 
-    echo "${C_BOLD}--- Rule Files ---${C_RESET}"
-    echo "Total .txt files compared: $total"
-    echo "Files with diffs:          $rule_diffs"
-    [ -n "$diff_list" ] && printf '%b\n' "$diff_list"
-    echo ""
-    echo "${C_BOLD}--- Added Files (informational) ---${C_RESET}"
-    echo "Files only on the changed side: $added_count  (new filters or platform targets, not a regression)"
```
maximtop, line 436: The Added Files section went away, but the loop that computes `added_count`/`added_list` just above still runs and its results now go nowhere. More importantly, files that exist only on the compare branch's side — a new filter or platform target, the usual case on a feature branch — no longer show up anywhere in the report, while the master-only case still fails as MISSING. Was dropping the section intentional? If yes, the loop and the locals should go with it; if not, restore it as informational.

scripts/build/strip-generated-meta.ts
```diff
@@ -64,11 +132,17 @@ const walkAndStrip = async (dir: string, rootDir: string): Promise<number> => {
         const fullPath = path.join(dir, entry.name);
 
         if (!entry.isDirectory()) {
-            // Only process .txt files that are directly inside a `filters/` dir,
-            // which is guaranteed by the caller when dir itself is a filters dir.
             if (entry.name.endsWith(TXT_FILE_EXTENSION)) {
                 return (await stripGeneratedMeta(fullPath)) ? 1 : 0;
             }
+            if (METADATA_FILE_NAMES.includes(entry.name)) {
```
maximtop, line 131: One thing this changes for the standalone CLI: `yarn strip-generated-meta` defaults to walking the tracked `platforms/` tree, so it now rewrites the 32 committed `filters.json`/`filters.js` in place, dropping fields the products read. A careless `git add -A` afterwards ships stripped metadata, and nothing in `yarn validate`/CI would catch it. Before, the same footgun only touched `!`-comment lines in `.txt`. Could the CLI require an explicit root, or at least warn when it touches tracked paths?

scripts/build/strip-generated-meta.ts
```diff
@@ -27,6 +34,66 @@ const isGeneratedMetaLine = (line: string): boolean => {
...
+};
+
+/**
+ * Strip generated version/timeUpdated fields from each entry of the `filters` array
+ * in a filters.json/filters.js metadata file, in-place.
+ *
+ * https://github.com/AdguardTeam/FiltersRegistry/issues/1208
+ *
+ * @param filePath - Absolute path to the filters.json or filters.js file.
+ * @returns True if the file was modified, false otherwise.
```
maximtop: nit: the doc promises a plain boolean, but invalid JSON throws and aborts the whole post-compile step. Worth an `@throws` here, and `@returns` should mention the skip branches, not just true/false.

scripts/compare-build-output-against-master.sh
```diff
@@ -371,22 +371,24 @@ generate_report() {
...
+  echo "filters.json/filters.js diffs: $meta_diffs"
+  echo ""
+  echo "${C_BOLD}--- Verdict ---${C_RESET}"
+  if [ "$rule_diffs" -eq 0 ] && [ "$meta_diffs" -eq 0 ]; then
+    echo "${C_GREEN}${CHECK} PASS${C_RESET} — no rule file or metadata diffs"
+  else
+    local fail_reasons=""
+    [ "$rule_diffs" -gt 0 ] && fail_reasons="$fail_reasons $rule_diffs rule file(s) differ"
+    [ "$meta_diffs" -gt 0 ] && fail_reasons="${fail_reasons:+$fail_reasons; }$meta_diffs metadata file(s) differ"
+    echo "${C_RED}${CROSS} FAIL${C_RESET} —$fail_reasons"
```
maximtop: Small rendering bug: in the metadata-only case this prints `FAIL —32 metadata file(s) differ` — the space after the dash exists only because the rule branch seeds `fail_reasons` with a leading space (line 387). Let the echo own the separator (`— $fail_reasons`) and drop the leading-space hack.

scripts/compare-build-output-against-master.sh
```diff
@@ -371,22 +371,24 @@ generate_report() {
...
-    [ -n "$added_list" ] && printf '%b\n' "$added_list"
-    echo ""
-    echo "${C_BOLD}--- Metadata Files (informational) ---${C_RESET}"
-    echo "filters.json/filters.js diffs: $meta_diffs  (version counter noise, not a regression)"
-    echo ""
-    echo "${C_BOLD}--- Verdict ---${C_RESET}"
-    [ "$rule_diffs" -eq 0 ] \
-        && echo "${C_GREEN}${CHECK} PASS${C_RESET} — no rule file diffs" \
-        || { echo "${C_RED}${CROSS} FAIL${C_RESET} — $rule_diffs rule file(s) differ"; return 1; }
+  echo "${C_BOLD}--- Rule Files ---${C_RESET}"
```
maximtop: nit: the rewritten block dropped to 2-space indent while the rest of the function (and the whole script) uses 4. Will fight every future edit of this region.

scripts/build/strip-generated-meta.ts
```diff
@@ -48,7 +115,8 @@ const stripGeneratedMeta = async (filePath: string): Promise<boolean> => {
 
 /**
  * Single-pass recursive walk: strips metadata from .txt files inside `filters/`
- * directories and counts modified files per `filters/` dir.
+ * directories, and from any `filters.json`/`filters.js` metadata files found
+ * anywhere under the root, counting modified files per `filters/` dir.
```
maximtop: This clause still says counting happens per `filters/` dir, but metadata files are now counted per file anywhere under the root (lines 138-145) — the `@returns` on `walkAndStrip` has the right semantics, the prose doesn't. (Different from the open scope thread on this doc.)

scripts/build/strip-generated-meta.ts
maximtop, line 131: nit: this branch nests a second `if` plus two `return 0` exits where the txt branch right above is a flat `return ... ? 1 : 0`. Same shape would read better here: capture the boolean, log when true, `return stripped ? 1 : 0;`.

scripts/compare-build-output-against-master.sh
```diff
@@ -371,22 +371,24 @@ generate_report() {
...
+  echo "${C_BOLD}--- Metadata Files ---${C_RESET}"
+  echo "filters.json/filters.js diffs: $meta_diffs"
+  echo ""
+  echo "${C_BOLD}--- Verdict ---${C_RESET}"
+  if [ "$rule_diffs" -eq 0 ] && [ "$meta_diffs" -eq 0 ]; then
+    echo "${C_GREEN}${CHECK} PASS${C_RESET} — no rule file or metadata diffs"
+  else
+    local fail_reasons=""
+    [ "$rule_diffs" -gt 0 ] && fail_reasons="$fail_reasons $rule_diffs rule file(s) differ"
+    [ "$meta_diffs" -gt 0 ] && fail_reasons="${fail_reasons:+$fail_reasons; }$meta_diffs metadata file(s) differ"
```
maximtop: The `${fail_reasons:+$fail_reasons; }` idiom plus the leading-space seed above makes this harder to read than the two-branch reality — and it already produced the dangling-semicolon bug earlier and the missing-space one below. There are at most two reasons; two plain `if ... echo` lines would be simpler.

scripts/build/__tests__/strip-generated-meta.test.ts
```diff
@@ -117,4 +144,141 @@ describe('stripGeneratedMetaFromDir', () => {
...
+        parsed.forEach((data) => {
+            expect(data.filters[0]).not.toHaveProperty('version');
+            expect(data.filters[0]).not.toHaveProperty('timeUpdated');
+            expect(data.filters[0].filterId).toBe(1);
+            expect(data.filters[0].name).toBe('Test filter');
+            expect(data.filters[0].timeAdded).toBe('2014-06-30T07:56:55+0000');
+        });
+    });
+
+    it('preserves tab indentation and trailing-newline from filters.json and filters.js', async () => {
```
maximtop, line 186: The name promises filters.js coverage, but the call falls back to the `['filters.json']` default, so the `.js` kind — the one with the riskiest strict-JSON assumption — has no byte-level round-trip pinned here. Pass both file names and assert both; the assertion loop already handles it.

scripts/build/strip-generated-meta.ts
```diff
@@ -27,6 +34,66 @@ const isGeneratedMetaLine = (line: string): boolean => {
     return GENERATED_META_PREFIXES.some((prefix) => line.startsWith(prefix));
 };
 
+/**
+ * Checks whether an array contains only non-null objects (i.e. records).
+ *
+ * @param value - The array to check.
+ * @returns True if every item in the array is a non-null object.
```
maximtop: nit: "non-null object" describes exactly the unsound guard we fixed in this branch — arrays are non-null objects too. Say "a plain object (non-null, not an array)".

scripts/build/strip-generated-meta.ts
```diff
@@ -64,11 +132,17 @@ const walkAndStrip = async (dir: string, rootDir: string): Promise<number> => {
...
         if (!entry.isDirectory()) {
-            // Only process .txt files that are directly inside a `filters/` dir,
-            // which is guaranteed by the caller when dir itself is a filters dir.
             if (entry.name.endsWith(TXT_FILE_EXTENSION)) {
                 return (await stripGeneratedMeta(fullPath)) ? 1 : 0;
             }
+            if (METADATA_FILE_NAMES.includes(entry.name)) {
+                if (await stripMetaFromMetadataFile(fullPath)) {
+                    // eslint-disable-next-line no-console
+                    console.log(`${path.relative(rootDir, fullPath)}: stripped generated meta fields`);
```
maximtop: nit: one run can now print three phrasings of the same message — `stripped generated meta fields` here, `stripped metadata from N file(s)` for `filters/` dirs, `Stripped generated meta from N new file(s).` in build.js. Pick one wording.

scripts/build/strip-generated-meta.ts
```diff
@@ -5,6 +5,7 @@ import { fileURLToPath } from 'url';
 
 const FILTERS_DIR_NAME = 'filters';
 const TXT_FILE_EXTENSION = '.txt';
+const METADATA_FILE_NAMES = ['filters.json', 'filters.js'];
```
maximtop: nit: the only list in this file without immutability typing — both siblings are `as const`. `const METADATA_FILE_NAMES: readonly string[] = [...]` says the same and keeps `.includes(entry.name)` working.

maximtop: This list and the compare script's `grep -cE 'filters\.(json|js)'` are two hand-maintained spellings of the same class (the docs make it three). They agree today, but add a name on only one side and either the gate silently skips it or the strip can't remove what the gate flags. Deriving the grep from this list — or a tiny agreement test — would keep them honest.

scripts/compare-build-output-against-master.sh
```diff
@@ -371,22 +371,24 @@ generate_report() {
...
-    [ "$rule_diffs" -eq 0 ] \
-        && echo "${C_GREEN}${CHECK} PASS${C_RESET} — no rule file diffs" \
-        || { echo "${C_RED}${CROSS} FAIL${C_RESET} — $rule_diffs rule file(s) differ"; return 1; }
+  echo "${C_BOLD}--- Rule Files ---${C_RESET}"
+  echo "Total .txt files compared: $total"
+  echo "Files with diffs:          $rule_diffs"
+  [ -n "$diff_list" ] && printf "$diff_list\n"
+  echo ""
+  echo "${C_BOLD}--- Metadata Files ---${C_RESET}"
+  echo "filters.json/filters.js diffs: $meta_diffs"
```
maximtop: Now that this count gates the verdict, the probe's imprecision is load-bearing: `diff -rq` also emits `Only in <dir>: filters.json` lines, which the grep counts as "differ" — an added/removed metadata file becomes a blocking FAIL with the wrong label. Meanwhile real diffs in `filters_i18n.json`/`local_script_rules.json` match neither `rule_diffs` nor `meta_diffs`, yet PASS prints "no rule file or metadata diffs". Filter the grep to `differ` lines at least; ideally also name what stays unchecked.

AGENTS.md
```diff
@@ -86,7 +86,7 @@ for all supported AdGuard products.
 | `yarn build:patches` | Build incremental patches |
 | `yarn generate-cache` | Generate cached `filter.txt` from templates (`tsx scripts/build/build.js --generate-cache`) |
 | `yarn download-stats` | Download per-filter `stats.json` to `temp/optimization/stats/` |
-| `yarn strip-generated-meta` | Strip generated meta lines from platform filter files |
+| `yarn strip-generated-meta` | Strip meta lines from `.txt` files + `version`/`timeUpdated` from `filters.{json,js}` |
```
maximtop: nit: "meta lines" and the `filters.{json,js}` brace shorthand appear only here — code, JSDoc and DEVELOPMENT.md say "generated meta" and `filters.json`/`filters.js`. Same term everywhere.

DEVELOPMENT.md
```diff
@@ -76,7 +76,8 @@ The `yarn build` command (and `yarn build:local`) also accepts:
   (`yarn build:patches`) is not needed afterwards.
 - `--strip-generated-meta` — after compilation, remove generated meta lines
   (`! Checksum`, `! Diff-Path`, `! TimeUpdated`, `! Version`) from all filter
-  files in `platforms/` and `temp/platforms/`. Useful when comparing outputs between builds.
+  files in `platforms/` and `temp/platforms/`, and remove the `version`/`timeUpdated`
```
maximtop, line 79: The "Command Compatibility" list below (line 177) still says the flag only removes metadata *lines* — now contradicts this updated bullet in the same file.

scripts/build/__tests__/strip-generated-meta.test.ts
```diff
@@ -93,22 +120,22 @@ describe('stripGeneratedMetaFromDir', () => {
             '',
         ].join('\n');
 
-        const oldFile = path.join(oldDir, 'test.txt');
-        const newFile = path.join(newDir, 'test.txt');
+        const oldFilePath = path.join(oldDir, 'test.txt');
```
maximtop, line 123: nit: the Dir/Path suffix convention stops one variable short — the fixture's `testFile` (line 16) is a full path without the suffix, right next to `oldFilePath`. Finish the rename or the churn bought nothing.

scripts/build/__tests__/strip-generated-meta.test.ts
```diff
@@ -117,4 +144,141 @@ describe('stripGeneratedMetaFromDir', () => {
...
+
+            const modified = await stripGeneratedMetaFromDir(platformsDir);
+            expect(modified).toBe(0);
+
+            const contentAfter = await fs.readFile(jsonFilePath, 'utf8');
+            expect(contentAfter).toBe(content);
+        }));
+    });
+
+    it('returns 0 and leaves the file untouched when filters array entries are null or arrays', async () => {
```
maximtop: The three "returns 0 and leaves the file untouched" tests run the identical six-line body; the first already uses a table-driven `cases` loop, the other two could fold into it so the skip contract lives in one place.

scripts/compare-build-output-against-master.sh
maximtop: Head still fails here — fbefaded folded `meta_diffs` into the verdict and dropped the `(informational)` labels, so every run against pre-merge master shows ~32 metadata diffs and exits 1. That contradicts the non-blocking-until-landed plan from your reply above; right now it breaks the compare workflow for every branch, this PR's own validation included. Keep it informational (or normalize the volatile fields out of the comparison) until master carries the strip.

scripts/build/strip-generated-meta.ts
mu-hun: This isn't a regression caused by the current PR. It's just stale code. I'm going to submit a separate PR to fix it.

mu-hun: Ah, It was my wrong guess. https://github.com/AdguardTeam/FiltersRegistry/pull/1226/changes#diff-e08a6035daa39b767eb8f8395c73a6b8b95e7e7eff609c2d67acb84867ffcce0

I'll fix them here.

mu-hun: Resolved in [`33bf686` (this PR)](https://github.com/AdguardTeam/FiltersRegistry/pull/1226/changes/33bf686b073065a01929c024bb72864842412d20).

scripts/compare-build-output-against-master.sh
mu-hun: This changes was a mistake from rebasing. I've reverted this change.

scripts/compare-build-output-against-master.sh
mu-hun: Applied in https://github.com/AdguardTeam/FiltersRegistry/pull/1226/commits/ea40dfd08326a23aa9cdac2dbb58216b7f1297b2

> Real diffs in filters_i18n.json/local_script_rules.json match nothing → PASS says "no diffs"

Following issue #1235, I misguidedly thought this shell script aligns the `filters/` directory with the `$BASE_BRANCH` baseline to keep `meta_diffs` at zero. I believed the metadata comparison step was unnecessary.

However, `filters*.json/.js` still contains `filterId`, `name`, `description`, `tags`, `homepage`, `deprecated`, `field` ordering, and formatting — All of which are generated by `generator.js`.  

A compiler update that changes how metadata is output (such as a new field, reordering, or a change in defaults) would only be caught during this step.

scripts/compare-build-output-against-master.sh
mu-hun: Reverted. This changes was a mistake from rebasing.

scripts/compare-build-output-against-master.sh
```diff
@@ -366,10 +366,12 @@ generate_report() {
...
     done < <(find "$PLATFORMS_CHANGED" -name "*.txt" -print0)
 
-    # Secondary check — metadata
+    # Secondary check — every generated non-rule file (filters*.json/.js,
+    # local_script_rules.json). Count only "Files X and Y differ"; "Only in ..."
+    # (present on one side only) is informational like added files above.
     meta_diffs=$(diff -rq --exclude="*.txt" --exclude="*.patch" \
         "$PLATFORMS_MASTER" "$PLATFORMS_CHANGED" \
-        | grep -cE "filters\.(json|js)" || true)
+        | grep -cE '^Files .* differ$' || true)
```
maximtop: [P3] Metadata-diff gate matches localized diff output — false PASS on non-English systems

The reworked probe counts metadata diffs by matching the English sentence that `diff -rq` prints: `grep -cE '^Files .* differ$'`. GNU diff (Linux, Git Bash on Windows) localizes that line through its message catalog — under `de_DE.UTF-8`, for example, it prints `Dateien <a> und <b> sind verschieden.`. The previous pattern (`grep -cE "filters\.(json|js)"`) matched the file name, which diff prints verbatim in every locale, so the old count was locale-independent.

Because `meta_diffs` now gates the verdict, a developer running `yarn compare-build-output` in a non-English locale gets `meta_diffs=0` regardless of real metadata diffs, and the script prints `PASS — no rule file or metadata diffs` and exits 0 — silently masking the exact class of divergence (filters.json/filters.js, filters_i18n, local_script_rules.json) that this gate was added to catch.

Suggest pinning the locale for this invocation, e.g. `LC_ALL=C diff -rq ...`, so the anchored pattern matches on every system.

maximtop (commented):
Independently validated findings from a multi-agent review. Only confirmed, non-duplicate issues are included.

mu-hun:
>  Only confirmed, non-duplicate issues are included.

What does the above mean? I didn't see any specific feedback in the code review to go off of.

maximtop:
> > Only confirmed, non-duplicate issues are included.
> 
> What does the above mean? I didn't see any specific feedback in the code review to go off of.

never mind, the agent went rogue

mu-hun:
I see. I'm waiting your additional reviews and approval 👀

slvvko (approved):

maximtop (approved):
No further comments from me. Ready to merge.

Alex-302:
> Ready to merge

after #1211?

mu-hun:
👀 Right. This PR will merge after down(parent) stack PR merged.

mu-hun:
This PR includes commits from #1212 due to a rebase a few weeks ago. I've squashed the commits by catalog, so please take another look when you have a chance.

Note: I discovered an issue where the metadata verdict is only PASSED when running `yarn build:local` in normal build mode.

```
--- Verdict ---
✗ FAIL
  34 metadata file(s) differ
```

mu-hun:
Fixed the `--use-cache` bug (filters.json/filters.js were silently omitted, see commit above).

One thing still shows up in the verdict — the "N metadata file(s) differ" — but that's not new, it's the same transitional mismatch I flagged earlier in https://github.com/AdguardTeam/FiltersRegistry/pull/1226#issuecomment-5392683137: master doesn't have this PR's `filters.json`/`filters.js` stripping yet, so it still carries `version`/`timeUpdated`.

To reproduce cleanly without that noise, run `git branch -f master HEAD` locally first (simulates post-merge state), then re-run `compare-build-output`.

105th (approved):
approving — the strip logic looks right, and the metadata gate asymmetry is already tracked in the thread above

maximtop (approved):

slvvko (approved):

Issue #1223, opened 2026-08-13
`--include`/`--skip` Filter ID Lists Are Silently Truncated on Windows


Alex-302:
## Summary

Running `yarn build --generate-cache --include=1,2,3 --skip=2` from PowerShell on
Windows compiles only filter 1 instead of filters 1 and 3. Unquoted commas are
converted to spaces when arguments are forwarded through the `yarn.cmd` shim, and
`parseFlags` in `scripts/build/build-config.ts` silently truncates the resulting
`"1 2 3"` value to `[1]` without any error or warning. The build finishes with
exit code 0, so the wrong result is indistinguishable from a successful build.

## Environment

- OS: Windows
- Shell: PowerShell 7.6.2 (pwsh, `PSNativeCommandArgumentPassing=Windows`) — affected
- Shell (control): Git Bash (MINGW64) — not affected
- Package manager: yarn 1.22.22
- Node.js: >= 22
- Repository: AdguardTeam/FiltersRegistry (master)

## Steps to Reproduce

1. Run `yarn build --generate-cache --include=1,2,3 --skip=2` in PowerShell.
2. Inspect the command echoed by yarn — the commas have been replaced with spaces:

    ```text
    tsx scripts/build/build.js --generate-cache "--include=1 2 3" --skip=2
    ```

3. Observe the build log: only filter 1 is compiled; filters 2 and 3 are skipped:

    ```text
    Filter 2 skipped due to '--include' option
    Filter 3 skipped due to '--include' option
    ```

## Expected Behavior

Filters 1 and 3 are compiled (their `filter.txt` and `revision.json` are updated);
filter 2 is skipped with the message `Filter 2 skipped due to '--skip' option`.

## Actual Behavior

Only filter 1 is compiled. Filters 2 and 3 are skipped with the message
`Filter X skipped due to '--include' option`, and the process exits with code 0
and no warning.

The message for filter 2 is the key evidence: if the include list had been parsed
as `[1, 2, 3]`, filter 2 would pass the include check and be rejected by the skip
check instead, producing `due to '--skip' option`. The fact that it was rejected
by the include check proves the parsed list contained only `[1]`.

## Root Cause

Two independent layers contribute to the bug.

### Layer 1: PowerShell argument serialization for batch files (environment)

- The environment under test is PowerShell 7.6.2 with
  `PSNativeCommandArgumentPassing=Windows` (legacy mode).
[...]

PR #1227, opened 2026-08-17
Fix/parse filter ids whitespace

cmondragon023:


piquark6046:
Why?

cmondragon023:
The change addresses issue #1223. On Windows PowerShell, an unquoted
--include=1,2,3 passed through the yarn.cmd shim can arrive as --include=1 2 3.
The previous parseInt() logic silently reduced that value to [1], producing a
successful but incomplete build.

This PR accepts both comma- and whitespace-separated IDs, rejects malformed IDs,
applies the same parser to patch generation, and adds regression tests plus
documentation.

scripts/build/filter-id-list.js
105th, line 1: make it ts, remove types from description

scripts/build/filter-id-list.js
```diff
@@ -0,0 +1,21 @@
...
+ * both comma- and whitespace-separated values. Reject malformed tokens instead
+ * of silently dropping them or partially parsing them with parseInt().
+ *
+ * @param {string} value - Raw value after the flag's equals sign.
+ * @param {string} flagName - Flag name used in validation errors.
+ * @returns {number[]} Parsed filter IDs.
+ */
+export function parseFilterIDs(value, flagName) {
+    const tokens = value.split(/[,\\s]+/).filter(Boolean);
+    const invalid = tokens.find((token) => !/^\\d+$/.test(token));
```
105th: instead of regexp use native functions as it was:
```
.map((x) => Number.parseInt(x, 10)).filter((x) => Number.isNaN(x));
```

scripts/build/filter-id-list.js
```diff
@@ -0,0 +1,21 @@
...
+ * PowerShell can replace unquoted commas when invoking a .cmd shim, so accept
+ * both comma- and whitespace-separated values. Reject malformed tokens instead
+ * of silently dropping them or partially parsing them with parseInt().
+ *
+ * @param {string} value - Raw value after the flag's equals sign.
+ * @param {string} flagName - Flag name used in validation errors.
+ * @returns {number[]} Parsed filter IDs.
+ */
+export function parseFilterIDs(value, flagName) {
+    const tokens = value.split(/[,\\s]+/).filter(Boolean);
```
105th: https://github.com/AdguardTeam/FiltersRegistry/pull/1227/changes#diff-a9e44feeea79afac2543b5465e4214066f9d01bc8e6dc7e0d8301965abe124ddR5

it will not split whitespace-separated values

Image

105th (changes requested):

cmondragon023:
Addressed the requested changes in the latest push: the shared parser is now TypeScript, JSDoc types were removed, parsing uses native string/number functions instead of regex, invalid IDs retain the previous filter-out behavior, and the whitespace examples are quoted and explained. Rebased onto current `master`; 23 focused tests, TypeScript typecheck, and ESLint all pass.

scripts/build/patches.js
```diff
@@ -32,15 +32,21 @@ args.forEach((val) => {
...
+            .filter((x) => !Number.isNaN(x));
     }
 
     if (val.startsWith('-s=') || val.startsWith('--skip=')) {
         const value = val.slice(val.indexOf('=') + 1);
 
         excludedFilterIDs = value
             .split(',')
-            .map((x) => Number.parseInt(x, 10));
+            .join(' ')
```
105th: why parseFilterIDs not used here?

105th (changes requested):

scripts/build/patches.js
cmondragon023: Good point — fixed in 39e0293d1. `patches.js` now reuses `parseFilterIDs`; `build:patches` runs through the existing `tsx` dependency so the JavaScript entry point can import the TypeScript helper. The full 50-test suite, typecheck, and ESLint pass.

cmondragon023:
All review threads are addressed in the current branch and have been marked resolved. Fresh verification on current `master`: 50/50 tests pass, TypeScript typecheck passes, and ESLint passes. @105th, could you please re-review when convenient?

cmondragon023:
Rebased onto the current `master` and addressed the remaining compatibility concern from the updated repository guidelines by migrating the touched patch builder from `patches.js` to `patches.ts`.

Fresh local verification on Node 22:

- `yarn lint`
- `yarn test` — 74 tests passed
- `yarn validate`
- `git diff --check upstream/master...HEAD`

The branch is now mergeable. @105th, could you please re-review when convenient?

mu-hun (commented):
Looks good on my end. How about adding a test case for the merged result of `--include=1,2,3 --skip=1,2` by extracting the code lines as a helper below. It's just my opinon, @105th might have a disagree with them.

https://github.com/AdguardTeam/FiltersRegistry/blob/3e069ec20ba2317f81510f8288310de28eb881ea/scripts/build/patches.ts#L95-L102

We are currently only validating the input case shown above.

https://github.com/AdguardTeam/FiltersRegistry/blob/5f26f5abce01a06d9d143ba781a9b7ccb941c89c/scripts/build/__tests__/build-config.test.ts#L168-L192

105th (approved):

mu-hun (approved):

Alex-302:
Thanks!

PR #20, opened 2018-02-17
fixed template for AlleBlock list

Alex-302:
https://github.com/AdguardTeam/FiltersRegistry/pull/19#issuecomment-366376926

PR #1237, opened 2026-09-04
AG-58576 Add update translations workflow

slvvko:
## Description

Task: AG-58576

Counterpart: [ext-compiler — add markdown log format to `validateLocales`](https://github.com/AdGuardSoftwareLimited/ext-compiler/pull/20)

### Locale validation error report

`yarn validate:locales` now uses the new markdown log format of `validateLocales` (ext-compiler PR above) and writes a `report_locales_error.md` file into the repo root when validation fails; stale reports are removed and the file is gitignored. The counterpart compiler PR is merged first; this PR then merges after that release and bumps `@adguard/filters-compiler` in `package.json`.

- `scripts/validation/validate_locales.ts` — pass `'markdown'` format; write/remove `report_locales_error.md`.
- `.gitignore` — ignore `report_locales_error.md`.

Example of the generated report:

```markdown
## Locales validation issues

### `bg`

- `critical` priority — **invalid or absent message key/value**:
  - `"filter.4.description": "Филтър за джаджите на социалните мрежи ( бутони "Харесвам" и подобни)."`

### `de`

- `critical` priority — **invalid or absent message key/value**:
  - `tag.3.description`

- `critical` priority — **invalid or absent message key/value**:
  - `tag.3.name`
```

### Translations download hardening

- `scripts/translations/download.sh` — `set -euo pipefail`, `SERVICE_URL` env override for testing, `curl --fail --retry 3 --retry-connrefused --connect-timeout 10 --max-time 120`.

### Validation comment reporting

- `scripts/translations/report-validation.ts` (new) — TypeScript module with a thin CLI that keeps the validation-failure comment on the translations PR in sync: it removes comments posted by previous runs and, on failure, posts a fresh comment with the latest report, truncated to the GitHub comment limit. Uses `@octokit/core` as the GitHub API client.
- `scripts/translations/__tests__/report-validation.test.ts` (new) — unit tests covering comment cleanup, report selection, truncation, and formatting.

### Weekly translations workflow

- `.github/workflows/update-translations.yaml` (new) — runs `download.sh` weekly (Mon 06:00 UTC) and on manual dispatch, validates locales (`yarn validate:locales`, `continue-on-error`), runs `yarn lint`/`yarn test` (the auto-created PR does not trigger the regular CI workflows), and opens a PR with the `locales/` changes; validation errors are reported as a PR comment via the module above.
- `AGENTS.md`, `DEVELOPMENT.md`, `scripts/translations/README.md` — docs.
- `@octokit/core` — new runtime dependency used by the reporting module.

## Testing Instructions

- `yarn validate:locales` — passes on the current `locales/` state; on failure it now produces `report_locales_error.md` and fails the command with a hint pointing to the report.
- `yarn test` — includes unit tests for the reporting module.
- Workflow can be triggered manually from the Actions tab.


slvvko:
/review considering it depends on the compiler's PR

scripts/validation/validate_locales.js
```diff
@@ -1,11 +1,16 @@
+import fs from 'fs';
```
maximtop: Please convert this script to TypeScript and update its package script and documentation references. AGENTS.md requires existing .js scripts to be migrated when touched; the new report-writing logic currently remains outside type checking because checkJs is false.

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,164 @@
...
+  workflow_dispatch:
+  schedule:
+    # Run every Monday at 06:00 UTC
+    - cron: '0 6 * * 1'
+
+concurrency:
+  group: update-translations
+  cancel-in-progress: false
+
+permissions:
```
maximtop: Could we scope these write permissions to the update job and give notify only the read permissions its Slack action needs? The workflow-level block currently gives the notification job contents, pull-requests, and issues write access, and that token is passed to the third-party Slack action.

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,164 @@
...
+              report = `... (truncated) ...\n${report.slice(-MAX_LOG_LENGTH)}`;
+            }
+            const body = [
+              MARKER,
+              '**`yarn validate:locales` failed** for the downloaded translations.',
+              'The PR was created anyway so the changes can be inspected and fixed.',
+              '',
+              '<details><summary>Validation report</summary>',
+              '',
+              report || 'No validation report was captured.',
```
maximtop: Please ensure translation values are safely encoded before embedding the validation report as Markdown. An invalid translation entry containing [demo](https://example.invalid) currently produces an active link in the bot comment; I verified this with GitHub's Markdown rendering API. The linked compiler PR #20 also needs a correction: backslash-escaping backticks inside a code span does not prevent a value containing backticks from closing that span and rendering a link outside it. Please fix this in the compiler's Markdown formatter, preserving the report structure, and verify the rendered output with such a value.

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,164 @@
...
+        with:
+          commit-message: 'skip ci: update translations'
+          add-paths: locales
+          branch: update-translations
+          base: master
+          title: Update translations
+          body: Translations are downloaded from Crowdin by the "Update translations" workflow.
+
+      - name: Report validation failure on the PR
+        if: steps.validate.outcome == 'failure'
```
maximtop: Please also clear or resolve the previous validation-failure comment when validation succeeds. For example: one run opens the translations PR with invalid data and posts a failure; a later run downloads corrected translations and updates that same PR successfully. This condition skips the entire cleanup step on the second run, leaving the old failure comment as the latest validation message.

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,164 @@
...
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 1
+
+      - name: Use Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
```
maximtop, line 32: Please add cache: 'yarn' to setup-node, as in the existing test and build workflows, so repeated translation updates can restore downloaded dependencies.

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,164 @@
...
+          branch: update-translations
+          base: master
+          title: Update translations
+          body: Translations are downloaded from Crowdin by the "Update translations" workflow.
+
+      - name: Report validation failure on the PR
+        if: steps.validate.outcome == 'failure'
+        uses: actions/github-script@v8
+        with:
+          script: |
```
maximtop: Could we move the report/comment logic into a module under scripts/ and keep this step as a thin caller? This block handles PR lookup, comment cleanup, report selection, truncation, and formatting; extracting it would let the normal lint/type checks cover it and make focused behavior tests easier to run. Since this is new script code, please use TypeScript in line with AGENTS.md.

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,164 @@
...
+            const fs = require('fs');
+            const path = require('path');
+            const { data: prs } = await github.rest.pulls.list({
```
maximtop: Could we give the create-pull-request step an id and use its pull-request-number output here? That would avoid rediscovering the same PR through pulls.list and repeating the branch name. Keep a guard for runs where the action does not return a PR number.

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,164 @@
...
+      - name: Validate locales
+        id: validate
+        continue-on-error: true
+        run: yarn validate:locales
+
+      # Creates or updates the update-translations branch and its pull request in a
+      # single step; does nothing when there are no translation changes. The branch
+      # is pushed and the PR is opened regardless of the validation result above.
+      - name: Create pull request
+        uses: peter-evans/create-pull-request@v6
```
maximtop: Optional: since this is a new workflow, consider using the current major versions of create-pull-request (v8) and github-script (v9), after checking their migration notes. This is a maintenance suggestion, not a blocker by itself.

maximtop (changes requested):
Please address the stale validation comment and translation-content escaping, narrow the notification job's permissions, and follow the repository's TypeScript migration rule. I also left requests to simplify the workflow and enable the Yarn cache; the action-version update is optional. The linked compiler PR was considered when checking the report formatting.

scripts/validation/validate_locales.js
slvvko: Fixed in e2e2e793: the script is converted to scripts/validation/validate_locales.ts (fully typed); the package script now runs the .ts file and AGENTS.md + scripts/translations/README.md references were updated.

.github/workflows/update-translations.yaml
slvvko: Fixed in 015d39a7: the workflow-level permissions block was removed; the update job declares contents/pull-requests/issues: write, and the notify job only contents: read (enough for its Slack action).

.github/workflows/update-translations.yaml
slvvko: Fixed in the counterpart compiler PR (AdGuardSoftwareLimited/ext-compiler#20, commit cd0a36e): escapeMarkdown now wraps translation content in code spans whose backtick delimiter run is one longer than the longest backtick run inside the value (backslash-escaping does not close code spans in GFM), and CRLF/CR/LF are collapsed to spaces. I verified the rendered output through GitHub's markdown API: the value with a backtick and a link stays inside a single code span, no active links, no injected headings. The report is embedded here exactly as the compiler generates it.

.github/workflows/update-translations.yaml
slvvko: Fixed in 33828889: the report step now runs on every run where the translations PR exists. On failure it deletes previous marker comments and posts a fresh report; on success it only deletes stale failure comments, so a fixed translation run leaves no outdated failure reminder on the PR.

.github/workflows/update-translations.yaml
slvvko: Fixed in 015d39a7: cache: 'yarn' added to setup-node, same as the test and build workflows.

.github/workflows/update-translations.yaml
slvvko: Fixed in 33828889: the logic moved to scripts/translations/report-validation.ts — a TypeScript module with a thin CLI and unit tests (stubbed fetch, no new dependencies); the workflow step is a thin caller. The module is covered by the regular lint/type checks and vitest (scripts/translations/__tests__/report-validation.test.ts).

.github/workflows/update-translations.yaml
slvvko: Fixed in 33828889: the create-pull-request step has id: create-pr and the report step uses steps.create-pr.outputs.pull-request-number (the action outputs it both on create and update); guarded by a condition for runs where the action does not return a PR number (no translation changes).

.github/workflows/update-translations.yaml
slvvko: Done in 015d39a7: create-pull-request bumped to v8 (migration notes checked — nothing relevant for our inputs). github-script is no longer used at all: its logic moved to the TypeScript module.

scripts/validation/validate_locales.ts
```diff
@@ -0,0 +1,73 @@
...
+// format is released and the dependency is bumped in package.json.
+type ValidateLocalesWithFormat = (
+    localesDirPath: string,
+    requiredLocales: string[],
+    logFormat: 'text' | 'markdown',
+) => ValidateLocalesResult;
+
+const validateLocalesWithFormat = validateLocales as ValidateLocalesWithFormat;
+
+const localesValidation = validateLocalesWithFormat(localesDirPath, REQUIRED_LOCALES, 'markdown');
```
105th, line 63: **High — Markdown log format is inert with the pinned compiler**

`validateLocalesWithFormat` is a blind cast, and the published `@adguard/filters-compiler@3.3.1` pinned in `package.json:35` accepts only `(localesDirPath, requiredLocales)` and ignores the third argument. The report written to `report_locales_error_<date>.md` is therefore the legacy text log, not markdown. I verified that 3.3.1 produces byte-identical output with and without `'markdown'`, and that the counterpart ext-compiler PR #20 is still open (npm `latest` remains 3.3.1). As merged, the PR description claim `now uses the new markdown log format` does not hold, and the cast will keep hiding any future signature drift.

Suggestion: merge this only after the compiler release (or pin the prerelease) and bump `@adguard/filters-compiler` in this PR; correct the PR description; and replace the blind cast with runtime feature detection so the fallback is explicit.

scripts/translations/report-validation.ts
```diff
@@ -0,0 +1,308 @@
...
+    const {
+        prNumber,
+        result,
+        repoRoot = process.cwd(),
+        client = buildDefaultClient(),
+    } = options;
+
+    const comments = await client.listComments(prNumber);
+    const staleComments = comments.filter((comment) => isValidationComment(comment.body));
+    await Promise.all(staleComments.map((comment) => client.deleteComment(comment.id)));
```
105th: **Medium — Cleanup runs before the new report and is not failure-isolated**

Stale comments are deleted via `Promise.all` before the replacement is posted, and the workflow step (`update-translations.yaml:72-76`) has no `continue-on-error`. On success, a single transient API error during cleanup fails the whole weekly job and sends a false Slack alert; on failure, an error in `createComment` after the deletes leaves the PR with no validation report at all. This partially defeats the intent of the earlier stale-comment fix.

Suggestion: use `Promise.allSettled` for the deletions, create the new comment before (or independently of) cleanup — or update the existing comment with a PATCH — and set `continue-on-error: true` on the report step; the dedicated `exit 1` step already provides the real failure signal.

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,106 @@
...
+        id: validate
+        continue-on-error: true
+        run: yarn validate:locales
+
+      # Creates or updates the update-translations branch and its pull request in a
+      # single step; does nothing when there are no translation changes. The branch
+      # is pushed and the PR is opened regardless of the validation result above.
+      - name: Create pull request
+        id: create-pr
+        uses: peter-evans/create-pull-request@v8
```
105th, line 71: **Medium — Auto-created PR will not trigger any CI checks**

`create-pull-request` uses the default `GITHUB_TOKEN`; push/pull-request events raised with that token do not start new workflow runs (GitHub platform behavior). The translations PR therefore gets no `test.yaml`/`build-adguard` checks, unlike human-opened translations PRs, and with branch protection the documented manual merge can be blocked by missing required checks. The change is only validated by the same job that produced it.

Suggestion: pass a PAT/App token to `create-pull-request` so the PR triggers CI, or run `yarn lint`/`yarn test` inside the `update` job; otherwise document that `validate:locales` is the only gate.

scripts/translations/report-validation.ts
```diff
@@ -0,0 +1,308 @@
...
+ * Report names embed a DD-MM-YYYY_HH-MM-SS timestamp, so sorting is
+ * chronological and the last entry is the most recent report.
+ *
+ * @param dir - Directory to scan for report files.
+ * @returns The report file name, or null when no report exists.
+ */
+export const findLatestReportFile = (dir: string): string | null => {
+    const reports = fs.readdirSync(dir)
+        .filter((name) => name.startsWith(REPORT_FILE_PREFIX) && name.endsWith(REPORT_FILE_EXTENSION))
+        .sort();
```
105th: **Medium — `findLatestReportFile` assumes `DD-MM-YYYY` names sort chronologically**

`.sort()` is lexicographic, but `formatDate` (`scripts/utils/strings.js:15`) emits `DD-MM-YYYY_HH-MM-SS`, so `report_locales_error_31-12-2026_...` sorts after `..._01-01-2027_...`: the `latest` pick is wrong across day/month boundaries. The docstring above (lines 77-78) and the test (`__tests__/report-validation.test.ts:79-91`, same-day fixtures only) encode the wrong assumption. It is latent only because `removeStaleReports()` normally leaves a single file.

Suggestion: sort by `fs.statSync(...).mtimeMs` or parse the timestamp into a `Date`; better, drop the `latest` logic and use a fixed report name since at most one report exists per run, and add a cross-boundary test.

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,106 @@
...
+    - cron: '0 6 * * 1'
+
+concurrency:
+  group: update-translations
+  cancel-in-progress: false
+
+jobs:
+  update:
+    name: Download translations and open a PR
+    runs-on: ubuntu-latest
```
105th, line 16: **Low — No timeouts on a long-running scheduled job**

The `update` job has no `timeout-minutes` (cf. `test.yaml:16`), and the 129 sequential `curl` calls in `scripts/translations/download.sh:57-59,92-94,125-127` have no `--connect-timeout`/`--max-time`/`--retry-connrefused`. With `concurrency.cancel-in-progress: false`, one stalled connection can hold the `update-translations` group for GitHub's 6-hour default and delay subsequent weekly runs.

Suggestion: add `timeout-minutes: 30` to the job and `--connect-timeout 10 --max-time 120 --retry-connrefused` to the curl calls.

105th (changes requested):
**Review summary**

Thanks for the update — the previously raised threads are fixed and resolved. A few issues remain before this can land, mainly around the compiler dependency and the reporting path:

- The markdown report format is not actually used: the pinned `@adguard/filters-compiler@3.3.1` ignores the third argument (the upstream PR is still open), so the report is the legacy text log and the `as` cast hides the mismatch.
- Comment reporting is not failure-isolated: deletions run before the new comment is posted and the step has no `continue-on-error`, so a transient API error can fail a successful run or leave the PR without a report.
- The auto-created PR will not trigger CI, because it is created with the default `GITHUB_TOKEN`.
- `findLatestReportFile` treats `DD-MM-YYYY` names as chronological, which is wrong across date boundaries.
- The scheduled job and the `curl` calls have no timeouts, so a stalled connection can hold the concurrency group for hours.

Please see the inline comments.

scripts/validation/validate_locales.ts
slvvko: Addressed. Per the agreed plan, the counterpart compiler PR is merged first; this PR merges after that release and bumps `@adguard/filters-compiler` in `package.json` — the PR description now states this explicitly. `validate_locales.ts` always passes `'markdown'`; the cast remains only until the bump and is marked with a TODO to drop. The report file also now has a fixed name (`report_locales_error.md`). See 366d632.

scripts/translations/report-validation.ts
slvvko: Fixed. The fresh report comment is now posted BEFORE stale-comment cleanup, so a transient API error during cleanup can no longer leave the PR without a report; cleanup uses `Promise.allSettled` (best-effort); the report step has `continue-on-error: true`, with the dedicated `exit 1` step remaining the real failure signal. Added a test covering the cleanup-failure path. See 8a11cc5 (module + tests) and e0b13d9 (workflow step).

.github/workflows/update-translations.yaml
slvvko: Addressed via the code-only option. The update job now runs `yarn lint && yarn test` (continue-on-error) and the final fail step also fails on lint/test failures, so the Slack notification fires. Documented in DEVELOPMENT.md and scripts/translations/README.md that `validate:locales` + lint + tests in the job are the only gate for the auto-created PR. Passing a PAT/App token is intentionally skipped for now. See e0b13d9 and 2a84b08.

scripts/translations/report-validation.ts
slvvko: Fixed with the preferred option: the latest-by-name logic is gone — the report file has a fixed name (`report_locales_error.md`), since `removeStaleReports()` guarantees at most one report per run, so cross-boundary sorting is structurally impossible. Tests updated; stale date-stamped files are ignored. See 366d632 and 8a11cc5.

.github/workflows/update-translations.yaml
slvvko: Fixed. `timeout-minutes: 30` on the update job and `--connect-timeout 10 --max-time 120 --retry-connrefused` on the three curl calls in `download.sh`. See e0b13d9.

scripts/validation/validate_locales.ts
```diff
@@ -0,0 +1,71 @@
...
+        .filter((name) => name.startsWith('report_locales_error') && name.endsWith('.md'));
+    staleReports.forEach((name) => fs.rmSync(path.join(repoRootPath, name), { force: true }));
+};
+
+// The `logFormat` argument is added by the counterpart compiler PR
+// (https://github.com/AdGuardSoftwareLimited/ext-compiler/pull/20), which is
+// merged and released before this PR. This PR bumps @adguard/filters-compiler
+// to that release; until the bump lands, the published types lack the argument,
+// hence the cast below.
+// FIXME: drop the cast after the dependency bump in package.json.
```
slvvko, line 54: FIXME

scripts/translations/report-validation.ts
```diff
@@ -0,0 +1,297 @@
...
+    if (result === 'failure') {
+        const report = readReport(repoRoot);
+        await client.createComment(prNumber, buildCommentBody(report));
+    }
+
+    // Cleanup is best-effort: Promise.allSettled swallows transient API errors
+    // (the workflow step also has continue-on-error), so a single failed
+    // deletion cannot fail an otherwise successful run.
+    const comments = await client.listComments(prNumber);
+    const staleComments = comments.filter((comment) => isValidationComment(comment.body));
```
105th: **Blocker — Fresh failure comment is deleted by the same run's cleanup**

On failure the fresh report comment is posted first (line 253) with `COMMENT_MARKER`, and then every comment whose body contains that marker is deleted (lines 259-261). `GET /issues/{n}/comments` returns the comment created moments earlier, so the newly posted comment is included in `staleComments` and deleted on every failure run — a validation failure leaves no comment at all, which is the feature this PR adds. The existing test masks this because its list mock omits the just-created comment.

Related (folded in): `listComments` is not failure-isolated despite the best-effort comment — it is awaited bare (line 259) — and rejected deletions are discarded with no logging (line 261).

Suggestion: delete stale comments before posting the fresh one (keeping `Promise.allSettled`), or make `createComment` return the created comment id and exclude it from deletion. Add a stateful-mock regression test (POST appends, GET returns all, DELETE removes) asserting exactly one marked comment remains after a failure run.

.gitignore
```diff
@@ -8,4 +8,7 @@ node_modules
 temp
 log.txt
 report_partial_*.txt
+report_locales_error_*.md
```
105th: **High — Ignore pattern does not match the generated report name**

`report_locales_error_*.md` requires an underscore before `.md`, but the fixed report name used by both `validate_locales.ts:16` and `report-validation.ts:17` is `report_locales_error.md` — `git check-ignore report_locales_error.md` exits 1. The PR claim that the file is gitignored, and the comment at `validate_locales.ts:13` (which also still cites the old `report_partial_*.txt` convention), are therefore false; a failed local validation leaves an untracked report that can be committed accidentally.

Additional impact: the workflow runs `yarn lint` in the same job, and `markdownlint **/*.md` picks up root-level files; the report is not excluded by `.markdownlintignore` (only `node_modules/`), so a generated report can fail the lint step and skip `yarn test`.

Suggestion: change the pattern to `report_locales_error.md` (or `report_locales_error*.md`), add the same entry to `.markdownlintignore`, and update the stale comment above the constant.

scripts/translations/__tests__/report-validation.test.ts
```diff
@@ -0,0 +1,273 @@
...
+    beforeEach(() => {
+        requestMock.mockReset();
+    });
+
+    const client = (): GitHubClient => new GitHubClient(
+        { request: requestMock } as unknown as Octokit,
+        'o/r',
+    );
+
+    it('posts a new failure comment and then removes stale comments', async () => {
```
105th: **Medium — Tests cannot catch the self-deletion bug and miss failure paths**

The list mock here returns only pre-existing comments, so it never models the comment created moments earlier by the POST — which is why the self-deletion in the cleanup path passes CI. Also missing: `listComments` rejecting after the failure comment is posted, a run with multiple stale comments, truncation boundaries (exactly 60000/60001), and any test for `parseArgs`. Temp dirs in this test and the placeholder test are removed without `finally`, leaking on assertion failure.

Suggestion: make the mock stateful (POST appends to a server-side list, GET returns the whole list, DELETE removes), then assert a failure run leaves exactly one marked comment and that the created id is never passed to DELETE. Add the missing failure/boundary cases and wrap cleanup in `finally`.

scripts/validation/validate_locales.ts
```diff
@@ -0,0 +1,71 @@
...
+    localesDirPath: string,
+    requiredLocales: string[],
+    logFormat: 'text' | 'markdown',
+) => ValidateLocalesResult;
+
+const validateLocalesWithFormat = validateLocales as ValidateLocalesWithFormat;
+
+const localesValidation = validateLocalesWithFormat(localesDirPath, REQUIRED_LOCALES, 'markdown');
+
+removeStaleReports();
```
105th, line 52: **Medium — Stale report can survive a thrown validation error**

`removeStaleReports()` runs only after `validateLocalesWithFormat()` returns (line 63), but the compiler throws when the locales dir is missing or empty (verified in the published `@adguard/filters-compiler@3.3.1`: `There is no locales dir` / `Locales dir ... is empty`) and can throw on other internal errors. On a throw, the previous run's fixed-name report remains at the repo root, and `readReport()` (`report-validation.ts:81-87`) performs no freshness check — so a stale report can be posted as the current failure, breaking the at-most-one-current-report invariant this design relies on.

Suggestion: call `removeStaleReports()` before validation, or wrap the validation call in `try/finally`, so no earlier report can outlive a thrown error.

105th (changes requested):
Requesting changes for three new issues (plus one folded in) found at this head — details in the threads:

- **Blocker**: the fresh validation-failure comment is deleted by the same run's stale-comment cleanup, so failures leave no comment at all.
- **High**: the `.gitignore` pattern does not match the fixed report name, and the generated report is not excluded from `markdownlint`; the "gitignored" claim is false.
- **Medium**: tests do not model the POST/GET/DELETE lifecycle, which is why the blocker passes CI.
- **Medium**: `removeStaleReports()` runs after validation, so a thrown validation error can leave a stale report that is then posted as current.

scripts/validation/validate_locales.ts
slvvko: Fixed in 8e7440a: the `ValidateLocalesWithFormat` cast and this FIXME are dropped — the published `@adguard/filters-compiler@3.3.2-beta.0` types include `logFormat?: LocalesLogFormat`, so the markdown format is actually used now.

scripts/translations/report-validation.ts
slvvko: Fixed in 6d4cc72: `createComment` now returns the created comment id, and `reportValidation` excludes it from stale-comment cleanup (`comment.id !== freshCommentId`), so a failure run leaves exactly one marked comment and the fresh id is never passed to DELETE. `listComments` is now failure-isolated (try/catch + `console.error`, cleanup skipped on failure) and rejected deletions are logged instead of being silently discarded. Added the requested stateful-mock regression test (POST appends / GET returns all / DELETE removes).

.gitignore
slvvko: Addressed: the `.gitignore` pattern is the exact file name `report_locales_error.md` (commit 66981d4); `report_locales_error.md` is now excluded via `.markdownlintignore` (commit 658ba24); and the stale comment above the constant in `validate_locales.ts` that cited the old `report_partial_*.txt` convention was rewritten (commit 8e7440a).

scripts/translations/__tests__/report-validation.test.ts
slvvko: Done in 4c5880a: the mock is now stateful (POST appends to a server-side list, GET returns the whole list including the just-created comment, DELETE removes) and the regression test asserts a failure run leaves exactly one marked comment and never passes the created id to DELETE. Added: `listComments` rejecting after the failure comment is posted, a run with multiple stale comments, truncation boundaries (exactly 60000/60001), and `parseArgs` (now exported). Temp dirs are cleaned up in `finally`.

scripts/validation/validate_locales.ts
slvvko: Fixed in 8e7440a: `removeStaleReports()` now runs before the validation call, so a thrown error (e.g. missing or empty locales dir) cannot leave a stale report at the repo root. Verified against the published `3.3.2-beta.0`: with the locales dir renamed, the run throws `There is no locales dir ...` and no report file remains afterwards.

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,124 @@
...
+        run: exit 1
+
+  notify:
+    name: Send Slack notification on failure
+    runs-on: ubuntu-latest
+    needs: update
+    # The notification job only reads repo data, so the default read-only token
+    # is enough; do not grant it the write permissions of the update job.
+    permissions:
+      contents: read
```
105th, line 112: should `actions: read` be added here? the `job` field in `fields` is resolved through the Actions API, and with an explicit `permissions` block every scope that isn't listed is `none`, so the lookup gets denied and the step fails before the webhook is sent. dropping `job` from `fields` would work too

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,124 @@
...
+    if:
+      ${{
+        always() &&
+        needs.update.result == 'failure' &&
```
105th: a run stopped by `timeout-minutes: 30` reports `cancelled`, not `failure`, so exactly the stalled-download case the timeout was added for skips the Slack alert here. `needs.update.result != 'success'` would cover it?

105th (changes requested):
a couple of notes on the notify job: the Slack step needs `actions: read` for the `job` field, and a run stopped by `timeout-minutes` reports `cancelled`, so it currently skips the alert.

maximtop (approved):
All comments are addressed. One thing before merge: bump @adguard/filters-compiler from 3.3.2-beta.0 to the stable 3.3.2 once it is out.

.github/workflows/update-translations.yaml
slvvko: Fixed in `5ba83647`: `actions: read` is added to the notify job's permissions so action-slack can resolve the `job` field through the Actions API; the field stays in `fields`, while the write scopes remain confined to the update job.

.github/workflows/update-translations.yaml
slvvko: Fixed in `6241ea7b`: the condition is now `needs.update.result != 'success'` (with `always()`), so a run stopped by `timeout-minutes` — reported as `cancelled` — fires the Slack alert too, which is exactly the stalled-download case the timeout protects against.

.github/workflows/update-translations.yaml
```diff
@@ -0,0 +1,130 @@
...
+        id: lint-test
+        continue-on-error: true
+        run: yarn lint && yarn test
+
+      # Creates or updates the update-translations branch and its pull request in a
+      # single step; does nothing when there are no translation changes. The branch
+      # is pushed and the PR is opened regardless of the validation result above.
+      - name: Create pull request
+        id: create-pr
+        uses: peter-evans/create-pull-request@v8
```
105th, line 71: a manual dispatch from a feature branch will check out that branch and create the `update-translations` PR from its HEAD into master. should we guard the job with `github.ref == 'refs/heads/master'`, like in `build-adguard.yaml`?

scripts/translations/report-validation.ts
```diff
@@ -0,0 +1,321 @@
...
+    body: string;
+}
+
+/**
+ * Checks whether a comment body was posted by this workflow.
+ *
+ * @param body - Comment body to check.
+ * @returns True when the body contains the workflow marker.
+ */
+export const isValidationComment = (body: string): boolean => body.includes(COMMENT_MARKER);
```
105th, line 39: the marker check matches anywhere in the body, and cleanup deletes by marker alone — a maintainer's quote reply copies the hidden marker, so their comment would be deleted on the next run. maybe also check `comment.user.type === 'Bot'`, or match the marker only at the start of the body?

105th (approved):
a couple of notes inline: a manual dispatch can build the auto-PR from a feature branch, and the marker-only cleanup can delete quoted replies. both look like easy follow-ups.

slvvko:
remaining unresolved review comments will be fixed in https://github.com/AdguardTeam/FiltersRegistry/pull/1252

PR #1244, opened 2026-09-15
fix: markdown indentation to 4-space from tab

mu-hun:
The editorconfig is missing this setting. Without this, VS Code sets the indentation to tabs.


piquark6046 (approved):
For me, it is OK

maximtop (approved):

PR #1252, opened 2026-09-18
AG-58576 address review follow-ups for the update-translations workflow

slvvko:
Follow-ups to #1237 (approved with notes):

- **Guard against feature-branch dispatch** (`github.ref == 'refs/heads/master'`, same pattern as `build-adguard.yaml`) — a manual dispatch from a feature branch can no longer create the `update-translations` PR from that branch's HEAD; the notify job now also excludes `skipped` runs so the guard does not raise a false Slack alert.
- **Protect maintainer quote replies from stale-comment cleanup** — `isValidationComment` now requires a bot author (`comment.user.type === 'Bot'`) and the marker at the very start of the body, with regression tests.
- **Pending:** bump `@adguard/filters-compiler` from `3.3.2-beta.0` to stable `3.3.2` once it is published on npm.


105th (approved):
both #1237 follow-ups look good — the master-only guard with the skipped-run exclusion, and the comment predicate with the regression tests. nothing to add

PR #1253, opened 2026-09-18
AG-58576 sync translations via Crowdin CLI instead of Twosky

slvvko:
## Description

The `Update translations` workflow fails with HTTP 403 from the Twosky gateway (`twosky.int.agrd.dev`) — Twosky blocks GitHub Actions runner IPs (verified: the same request works from a local machine but returns 403 from runners; see run [35374101307](https://github.com/AdguardTeam/FiltersRegistry/actions/runs/35374101307)).

The underlying translations live in the Crowdin project `adguard-applications` (project ID 17570, folder `miscellaneous/filters-registry`) — Twosky is only an internal proxy for it. This PR switches the download to the Crowdin CLI.

**This is the same approach the KnowledgeBase repos use** ([example workflow](https://github.com/AdguardTeam/KnowledgeBase/blob/master/.github/workflows/crowdin.yml)): `crowdin.yml` with `project_id` + `api_token_env`, `@crowdin/cli` as a dependency, and the `CROWDIN_PERSONAL_TOKEN` secret passed in the workflow. Unlike the KnowledgeBase flow (`crowdin upload && crowdin download --export-only-approved`), here upload stays a manual step (see below), and the CLI is used for download only.

### Changes

- **`crowdin.yml`** (new) — project 17570 config: `export_languages` (42 locales matching the old `download.sh` list), `languages_mapping` (`zh-CN` → `zh`, `pt-BR`/`pt-PT` → underscore dirs), `dest` pointing at the existing `miscellaneous/filters-registry` folder.
- **`scripts/translations/download-crowdin.sh`** (new) — downloads all translations via the Crowdin CLI and converts them into `locales/` with the existing `converter.js`, preserving the `es` → `es_ES` / `pt` → `pt_PT` copy behavior.
- **`.github/workflows/update-translations.yaml`** — runs `download-crowdin.sh` with `CROWDIN_PERSONAL_TOKEN` from repo secrets.
- **`package.json` / `yarn.lock`** — `@crowdin/cli` devDependency + `crowdin` script.
- **Docs** — `scripts/translations/README.md`, `DEVELOPMENT.md`, `AGENTS.md`.

### Upload stays manual

Uploading strings to the service remains a **manual-only** step via the legacy `upload.sh` (Twosky, run from an allowlisted network). There is intentionally no Crowdin-CLI-based upload script — documented in `AGENTS.md`.

## Prerequisites before merging

1. Add the **`CROWDIN_PERSONAL_TOKEN`** secret to the repo (Settings → Secrets and variables → Actions). The token needs access to the `adguard-applications` Crowdin project — the same token used by the KnowledgeBase repos works.
2. Then trigger the workflow manually (Actions → Update translations → Run workflow) to verify it opens the `update-translations` PR.

## Testing

- `yarn lint:code`, `yarn lint:types` — pass.
- `yarn test` — 80/80 tests pass (one suite fails to *load* on this machine due to a local `yarn link` to `/Volumes/dev/COPY/ext-compiler`; identical on clean `master`, unrelated).
- `crowdin download --dryrun` with the config parses correctly (fails only on the dummy token, as expected).


scripts/translations/download-crowdin.sh
```diff
@@ -0,0 +1,58 @@
...
+locales=(
+    "ar" "be" "bg" "ca" "cs" "da" "de" "el" "es"
+    "fa" "fi" "fr" "he" "hi" "hr" "hu" "hy" "id" "it"
+    "ja" "ko" "lt" "mk" "ms" "nl" "no" "pl" "pt" "pt_BR"
+    "pt_PT" "ro" "ru" "sk" "sl" "sr" "sv" "th" "tr" "uk"
+    "vi" "zh" "zh_TW"
+)
+
+echo "Downloading translations"
+yarn -s crowdin download --config "$crowdinConfig"
```
slvvko: The crowdin CLI resolves the config's `source:` patterns against local files, and `crowdin.yml` points them at `temp/crowdin/source/*.json`, which nothing in this PR ever creates (there's no upload step). Without `--all` the CLI throws "Couldn't find any file to download" — or, if skip-untranslated-files is enabled on the project, exits 0 having downloaded nothing — so the workflow can't produce translations either way. The `--dryrun` check can't catch this: it fails at auth before the file mapping is resolved. I think we should pass `--all` here and validate the archive layout with a real token.

crowdin.yml
```diff
@@ -0,0 +1,76 @@
+project_id: "17570"
+api_token_env: "CROWDIN_PERSONAL_TOKEN"
+preserve_hierarchy: true
+
+# Keep the download limited to the locales this repo tracks; the codes are the
+# same ones the Twosky-based download.sh used.
```
slvvko: We're dropping `en` from the download set: the old `download.sh` fetched 43 locales including `en`, these lists have 42 without it. `locales/en` is committed, is required by `validate_locales`, and is what `upload.sh` pushes to the service — if it's never refreshed it goes stale silently, and the comment "the codes are the same ones the Twosky-based download.sh used" stops being accurate. `en` is a valid Crowdin language code; I'd add it to `export_languages` and to the script's `locales` array.

scripts/translations/download-crowdin.sh
```diff
@@ -0,0 +1,58 @@
...
+
+for locale in "${locales[@]}"; do
+    echo "Importing $locale locale"
+    for spec in "tags.json:tag." "groups.json:group." "filters.json:filter."; do
+        file="${spec%%:*}"
+        mask="${spec##*:}"
+        if [ ! -f "$crowdinDir/$locale/$file" ]; then
```
slvvko, line 40: If a file isn't downloaded the loop prints "Skip" and continues, and the script still exits 0. The old script failed loudly on any missing file (`curl --fail` under `set -e`), so a broken or partial download now produces a green job with stale translations, no PR, and no alert — `validate_locales` passes on the old files because they're still valid JSON. Could we count imported/skipped files, fail when nothing was imported at all, and print the skip summary so the workflow's failure path still fires?

crowdin.yml
```diff
@@ -0,0 +1,76 @@
...
+export_languages:
```
slvvko, line 9: Some of these codes are Twosky-style and don't exist in Crowdin: bare `es` (the ID is `es-ES`), `pt` (`pt-PT`/`pt-BR`), `sv` (`sv-SE`), `hy` (`hy-AM`) — I checked against the Crowdin language list. The CLI validates every `export_languages` entry against the project's languages and aborts the whole download on the first mismatch. I'd switch to the Crowdin IDs and map them to the repo dirs in `languages_mapping` (e.g. `sv-SE: sv`, `hy-AM: hy`); the `es-ES: es_ES` mapping then becomes real instead of defensive.

scripts/translations/download-crowdin.sh
```diff
@@ -0,0 +1,58 @@
...
+        fi
+        node converter.js import "$crowdinDir/$locale/$file" "$locale" converted.json "$mask"
+        mkdir -p "$workDir/locales/$locale"
+        cp -f converted.json "$workDir/locales/$locale/$file"
+        rm converted.json
+    done
+
+    # es_ES is a copy of es; pt_PT also receives pt's tags and groups as a
+    # fallback when the pt-PT language has no own download (mirrors download.sh).
+    if [ "$locale" = "es" ]; then
```
slvvko: These fallback blocks are order- and presence-dependent: the pt→pt_PT copy only works because `pt` happens to precede `pt_PT` in the array (reordering would overwrite a genuine download), the copies run even when the source locale was skipped (stale data), and if `es-ES` is ever exported the fresh `es_ES` files would be silently clobbered by the `es` copy. Once the languages are exported under their real Crowdin IDs (`es-ES`, `pt-PT`), the fallbacks become unnecessary — I'd drop the copy blocks and download `es_ES`/`pt_PT` directly.

scripts/translations/download-crowdin.sh
```diff
@@ -0,0 +1,58 @@
...
+        mkdir -p "$workDir/locales/es_ES"
+        for file in tags.json groups.json filters.json; do
+            cp -f "$workDir/locales/es/$file" "$workDir/locales/es_ES/$file"
+        done
+    fi
+
+    if [ "$locale" = "pt" ]; then
+        echo "Copying pt translations to pt_PT"
+        mkdir -p "$workDir/locales/pt_PT"
+        for file in tags.json groups.json; do
```
slvvko: The es→es_ES fallback copies all three files, but the pt→pt_PT fallback copies only tags and groups, so `pt_PT/filters.json` can sit stale with no signal when pt-PT has no own download. If we export pt-PT directly (per the comment on `export_languages`) this whole fallback goes away — worth aligning the two while we're here.

scripts/translations/download-crowdin.sh
```diff
@@ -0,0 +1,58 @@
...
+# after the languages_mapping in crowdin.yml (zh-CN is stored as `zh`)
+locales=(
+    "ar" "be" "bg" "ca" "cs" "da" "de" "el" "es"
+    "fa" "fi" "fr" "he" "hi" "hr" "hu" "hy" "id" "it"
+    "ja" "ko" "lt" "mk" "ms" "nl" "no" "pl" "pt" "pt_BR"
+    "pt_PT" "ro" "ru" "sk" "sl" "sr" "sv" "th" "tr" "uk"
+    "vi" "zh" "zh_TW"
+)
+
+echo "Downloading translations"
```
slvvko, line 27: nit: `temp/crowdin` is never cleaned between runs, so a file the current run no longer produces (removed from the project, locale dropped from `export_languages`) stays in the dir from the previous run and gets re-converted into `locales/` — stale translations kept alive, meaningless PRs. `rm -rf` it right before the download so the `[ -f ]` checks below actually mean something.

scripts/translations/download-crowdin.sh
slvvko: Fixed in e0aa0338ec33 — the download now passes `--all`, so the CLI matches the configured files against the server-side project by `dest` instead of the local source paths.

crowdin.yml
slvvko: Fixed in cb45000f8170 — `en` added back to `export_languages` and to the script's `locales` array.

scripts/translations/download-crowdin.sh
slvvko: Fixed in 9833a2c9d8bc — the script now counts imported/skipped files, exits 1 with an error when nothing was imported, and prints the skip summary at the end.

crowdin.yml
slvvko: Fixed in 249ba7d1cd10 — codes switched to `es-ES`/`pt-PT`/`sv-SE`/`hy-AM` (all 42 verified against the Crowdin language list), with `sv-SE: sv` and `hy-AM: hy` added to `languages_mapping`.

scripts/translations/download-crowdin.sh
slvvko: Fixed in f1667f3baa7d — the copy blocks are removed; `es_ES` and `pt_PT` are now downloaded directly under their Crowdin IDs.

scripts/translations/download-crowdin.sh
slvvko: Fixed in the same commit as the fallback removal (f1667f3baa7d) — with `pt-PT` exported directly, `pt_PT/filters.json` comes from the download, so the fallback gap is gone.

scripts/translations/download-crowdin.sh
slvvko: Fixed in b4cfb0699d1a — `temp/crowdin` is now cleaned with `rm -rf` right before the download, so the `[ -f ]` checks only see files from the current run.

scripts/translations/download-crowdin.sh
```diff
@@ -0,0 +1,58 @@
...
+# Twosky gateway. Requires the CROWDIN_PERSONAL_TOKEN environment variable
+# (see crowdin.yml api_token_env). Run from the scripts/translations dir.
+# Uploading strings to the service is a manual step done via upload.sh.
+workDir=../..
+crowdinDir="$workDir/temp/crowdin"
+crowdinConfig="$workDir/crowdin.yml"
+
+# keep the list alphabetically sorted; the dir names are the repo locale dirs
+# after the languages_mapping in crowdin.yml (zh-CN is stored as `zh`)
+locales=(
```
maximtop: I'd keep `locales/es` and `locales/pt` updated too. The old script refreshed both, and they are still required by `validate_locales.ts` and consumed by the compiler. This loop only writes `es_ES` and `pt_PT`, so a successful download leaves the base locales permanently stale. Copying the freshly imported regional files back to their base aliases would preserve those outputs without restoring the old overwrite problem.

crowdin.yml
```diff
@@ -0,0 +1,79 @@
+project_id: "17570"
+api_token_env: "CROWDIN_PERSONAL_TOKEN"
+preserve_hierarchy: true
+
+# Keep the download limited to the locales this repo tracks. The codes are
+# Crowdin language IDs; the Twosky codes es/pt/sv/hy are es-ES/pt-PT/sv-SE/
+# hy-AM in Crowdin (the CLI hard-fails on codes the project does not have).
+# `en` is downloaded like in the old download.sh.
+export_languages:
```
maximtop, line 13: nit: I'd use this configuration as the single source for the import locale list. `download-crowdin.sh` independently repeats both the language set and its mapped directory names, so adding a language here downloads it but silently skips importing it, even reporting zero skipped files. Deriving the import directories from the config or the fresh download would avoid that drift.

maximtop (changes requested):
I'd preserve updates to the base Spanish and Portuguese locales before merging.

scripts/translations/download-crowdin.sh
slvvko: Fixed in a4d5b4a1b886: `download-crowdin.sh` now refreshes the base aliases after the import — `es_ES` → `es` and `pt_PT` → `pt`, all three files each, so the `pt_PT/filters.json` gap of the old Twosky flow is closed as well.

The copies run in a separate pass after the import (no order dependence), are skipped when the alias dir was downloaded on its own (a future `es`/`pt` export cannot be clobbered by the alias copy), and fail with an explicit error when the source files were not imported instead of copying stale data.

`es`/`pt` are also documented next to `languages_mapping` in `crowdin.yml` and in `scripts/translations/README.md`.

crowdin.yml
slvvko: Fixed in e9041abe7fce: the hardcoded `locales` array is gone. The script now imports exactly the locale dirs the download produced (`find temp/crowdin -maxdepth 1 -type d`, excluding the config `source` dir), so `export_languages` + `languages_mapping` are the single source for both the language set and the mapped dir names — a language added there is downloaded and imported.

To keep the reporting that a derived list loses, a downloaded dir with none of the expected files is now a hard error, and repo locale dirs that the run did not refresh are reported as a warning (that is what catches e.g. `en` going stale).

scripts/translations/download-crowdin.sh
```diff
@@ -0,0 +1,151 @@
...
+# produces (removed from the project, locale dropped) would otherwise stay
+# in temp/crowdin and be re-imported, keeping stale translations alive.
+echo "Cleaning previous downloads"
+rm -rf "$crowdinDir"
+
+echo "Downloading translations"
+# --all makes the CLI match the configured files against the server-side
+# project files (by dest) instead of the local source paths, which are not
+# staged in this repo (there is no upload step).
+yarn -s crowdin download --all --config "$crowdinConfig"
```
maximtop, line 37: `yarn run` executes scripts from the repo root, so `--config ../../crowdin.yml` points two levels above the repo. The CLI fails with `File '../../crowdin.yml' does not exist` (exit 102), and the workflow never downloads anything. The command from the README fails the same way.

I'd resolve the config from the script dir, e.g. `"$(cd "$(dirname "$0")/../.." && pwd)/crowdin.yml"`, or pass `--config crowdin.yml`, which yarn resolves against the root.

scripts/translations/download-crowdin.sh
```diff
@@ -0,0 +1,151 @@
...
+baseLocaleAliases=(
+    "es_ES:es"
+    "pt_PT:pt"
+)
+
+# Clean up the previous download first: a file the current run no longer
+# produces (removed from the project, locale dropped) would otherwise stay
+# in temp/crowdin and be re-imported, keeping stale translations alive.
+echo "Cleaning previous downloads"
+rm -rf "$crowdinDir"
```
maximtop, line 31: nit: `workDir=../..` is relative to the caller's cwd, so running the script from anywhere but `scripts/translations` makes this `rm -rf` hit `../../temp/crowdin` outside the repo. Deriving `workDir` from `dirname "$0"` would fix this together with the config path below.

scripts/translations/download-crowdin.sh
```diff
@@ -0,0 +1,151 @@
...
+# their dirs are named, so the config stays the single source for the language
+# set and this script needs no locale list of its own: a language added there is
+# downloaded and imported, and a dropped one stops being imported.
+# (`source` is the config's source dir, not a locale.)
+while IFS= read -r locale; do
+    echo "Importing $locale locale"
+    localeImported=0
+    for i in "${!files[@]}"; do
+        file="${files[$i]}"
+        if [ ! -f "$crowdinDir/$locale/$file" ]; then
```
maximtop, line 62: A single missing file is only logged as `Skip`, and the job stays green. The old script failed on any missing file. Combined with the alias copy below, a stale committed `es_ES` file can also end up in `es`. Unless Crowdin skips untranslated files on purpose, I'd fail when a downloaded locale dir is missing some of the expected files.

scripts/translations/download-crowdin.sh
```diff
@@ -0,0 +1,151 @@
...
+# Twosky gateway. Requires the CROWDIN_PERSONAL_TOKEN environment variable
+# (see crowdin.yml api_token_env). Run from the scripts/translations dir.
+# Uploading strings to the service is a manual step done via upload.sh.
+workDir=../..
+crowdinDir="$workDir/temp/crowdin"
+crowdinConfig="$workDir/crowdin.yml"
+
+# The files exported per locale and the converter.js mask for each of them.
+# Keep in sync with the `files:` section of crowdin.yml.
+files=("tags.json" "groups.json" "filters.json")
```
maximtop, line 13: nit: `files`/`masks` repeat the `files:` section of `crowdin.yml`. A file added there is downloaded but silently not imported. I'd fail when a downloaded locale dir contains a file that isn't in this list, so the drift shows up.

package.json
```diff
@@ -21,6 +21,7 @@
         "validate": "yarn validate:platforms && yarn validate:locales",
         "validate:platforms": "tsx scripts/validation/validate_platforms.js",
         "validate:locales": "tsx scripts/validation/validate_locales.ts",
+        "crowdin": "crowdin",
```
maximtop, line 24: nit: the new `crowdin` script is missing from the commands table in `AGENTS.md`.

AGENTS.md
```diff
@@ -33,7 +33,10 @@ for all supported AdGuard products.
     - `build-adguard.yaml`
     - `build-3p.yaml`
     - `test.yaml` (runs `yarn test` on every pull request)
-    - `update-translations.yaml` (weekly translations download that opens a PR)
+    - `update-translations.yaml` (weekly translations download from Crowdin that opens a PR;
+      requires the `CROWDIN_PERSONAL_TOKEN` secret)
+    - Uploading strings to the translation service is **manual only**, via the legacy
```
maximtop, line 38: nit: this bullet isn't a workflow, but it sits in the CI workflows list. I'd move it out of the list.

maximtop (changes requested):
The download step can't find `crowdin.yml` when run through `yarn`, so the workflow fails before downloading anything. Details inline.

scripts/translations/download-crowdin.sh
maximtop: The guard in `download-crowdin.sh:112` checks the committed `locales/es_ES`/`pt_PT` files, which always exist in a checkout, so the "was not imported" error never fires. If `es_ES/<file>` is missing from this download, the stale committed file is copied to `es` without any signal. I'd check against what this run actually imported instead of `locales/`.

105th (changes requested):
can we make the config path absolute? yarn runs package scripts from the repo root, so `--config ../../crowdin.yml` points outside the repo and the download aborts before anything is fetched

scripts/translations/download-crowdin.sh
slvvko: Fixed in 398afaa7f131: `crowdinConfig` is now derived from the script's own location — `workDir="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"` — so `--config` is an absolute path that exists regardless of the caller's cwd (`yarn run` from the repo root, CI from `scripts/translations`, anywhere else). Verified with a stub CLI that exits 102 when `--config` does not exist: it never fires. This also covers 105th's request.

scripts/translations/download-crowdin.sh
slvvko: Fixed in 398afaa7f131: `workDir` is resolved from the script dir, so `rm -rf "$crowdinDir"` always targets `<repo>/temp/crowdin` and can never hit a path outside the repository. `converter.js` paths stay relative to `scripts/translations` (it resolves them with `path.join(__dirname, …)`), and the script `cd`s into its own dir for the `converted.json` scratch file.

scripts/translations/download-crowdin.sh
slvvko: Fixed in 9ba833111f7a: a downloaded locale dir missing any of the three expected files is now a hard error (`Error: <locale>/<file> was not downloaded …`), matching the old `download.sh`'s `curl --fail` behaviour. With `--all` the CLI writes all mapped files for every exported language, so a missing file is a real anomaly, not a routine skip; the `skipped` counter is gone and the summary is now `Import finished: <N> files imported, <M> base locales refreshed`.

scripts/translations/download-crowdin.sh
slvvko: Fixed in 9ba833111f7a: the import loop now fails when a downloaded locale dir contains a file that is not in the `files` list (`Error: unexpected file(s) in the downloaded <locale> dir: …`), so adding a file to the `files:` section of `crowdin.yml` without updating this script breaks the run instead of silently dropping the import. `files`/`masks` remain the single source for the converter masks (not derivable from YAML without a parser).

package.json
slvvko: Fixed in d222bec713bb: `yarn crowdin` added to the commands table in `AGENTS.md`.

AGENTS.md
slvvko: Fixed in d222bec713bb: the "upload is manual only" bullet is moved out of the CI workflows list into its own top-level bullet under Technical Context.

scripts/translations/download-crowdin.sh
slvvko: Fixed in 5b84a1ebe94b: the alias refresh now guards on what this run actually imported (`importedLocales`, which after the per-file fix implies all three files) instead of the committed `locales/` dirs that always exist. Reproduced the exact scenario — committed `es_ES` present, only `ar` downloaded — the run now fails with `cannot refresh the es locale: es_ES was not imported by this run` instead of copying stale files into `es`.

maximtop (approved):
LGTM

105th (approved):
config path fix resolves my comment :+1:

